PR #27, opened 2024-03-15
add `qeye()=eye()`

ytdHuang:
Support `qeye` which makes the syntax more similar to `QuTiP`.


ytdHuang:
Something wrong, will make another PR later

PR #29, opened 2024-03-18
support `Base.eltype` for `QuantumObject`

ytdHuang:
This helps to directly obtain the element type in `QuantumObject.data`, e.g., `Int64`, `Float64`, `ComplexF64` ...
I think this would be quite useful when developing new functionality in the future.

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -19,6 +19,7 @@ isket
 isoper
 issuper
 size
+eltype
 length
 LinearAlgebra.tr
 LinearAlgebra.norm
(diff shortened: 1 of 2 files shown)


ytdHuang:
I'm didn't write a test for this feature since it's just calling the function in `Base`.
For the code coverage, it will be automatically covered by the tests for new developed features.

albertomercurio:
Yes this is fine. Thank you.

BTW, I'm at RIKEN right now 😄

ytdHuang:
> Yes this is fine. Thank you.
> 
> BTW, I'm at RIKEN right now 😄

Nice~ I'll arrive at Sunday afternoon this week.

PR #30, opened 2024-03-18
Calculate Negativity

ytdHuang:
1. New exported functions:
    - [ ] `negativity`
    - [ ] `partial_transpose` (for either dense or sparse `QuantumOpjects`, but only allow `type=OperatorQuantumObject`)

2. New runtest set for above functions

### DIFF
--- a/src/negativity.jl
+++ b/src/negativity.jl
@@ -1,0 +1,30 @@
+@doc raw"""
+    partial_transpose(ρ::QuantumObject, mask::Vector{Bool})
+
+Return the partial transpose of a density matrix ``\rho``, where `mask` is an array/vector with length that equals the length of `ρ.dims`. The elements in `mask` are boolean (`true` or `false`) which indicates whether or not the corresponding subsystem should be transposed.
+"""
+function partial_transpose(ρ::QuantumObject{::T, OperatorQuantumObject}, mask::Vector{Bool}) where T
+    if length(mask) != length(ρ.dims) 
+        error("The length of \`mask\` should be equal to the length of \`ρ.dims\`.")
+    end
+    return _partial_transpose(ρ, mask)
+end
+
+function _partial_transpose(ρ::QuantumObject{<:AbstractMatrix, OperatorQuantumObject}, mask::Vector{Bool})
+    mask2 = [1 + Int(i) for i in mask]
+    # mask2 has elements with values equal to 1 or 2
+    # 1 - the subsystem don't need to be transposed
+    # 2 - the subsystem need be transposed
+
+    nsys = length(mask2)
+    pt_dims = reshape(Vector(1:(2 * nsys)), (nsys, 2))
+    pt_idx  = [
+        [pt_dims[n,     mask2[n]] for n in 1:nsys]; # origin   value in mask2
+        [pt_dims[n, 3 - mask2[n]] for n in 1:nsys]  # opposite value in mask2 (1 -> 2, and 2 -> 1)
+    ]
+    return QuantumObject(
+        reshape(permutedims(reshape(ρ.data, (ρ.dims..., ρ.dims...)), pt_idx), size(ρ)),
+        OperatorQuantumObject,
+        ρ.dims
+    )
+end
--- a/src/negativity.jl
+++ b/src/negativity.jl
@@ -1,7 +1,9 @@
 @doc raw"""
     negativity(ρ::QuantumObject, subsys::Int; logarithmic::Bool=false)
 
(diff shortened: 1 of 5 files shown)


albertomercurio (commented):
Wonderful. I was wondering if it were possible to make it agnostic on the array tipe, thus supporting also CUDA and other arrays.

src/negativity.jl
```diff
@@ -0,0 +1,136 @@
...
+    #   2 - the subsystem need be transposed
+
+    nsys = length(mask2)
+    pt_dims = reshape(Vector(1:(2 * nsys)), (nsys, 2))
+    pt_idx  = [
+        [pt_dims[n,     mask2[n]] for n in 1:nsys]; # origin   value in mask2
+        [pt_dims[n, 3 - mask2[n]] for n in 1:nsys]  # opposite value in mask2 (1 -> 2, and 2 -> 1)
+    ]
+    return QuantumObject(
+        reshape(permutedims(reshape(ρ.data, (ρ.dims..., ρ.dims...)), pt_idx), size(ρ)),
```
albertomercurio, line 90: Be aware that I have already implemented the `row_major_reshape` function, if you need it. Because it seems that what you are doing here is very similar.

ytdHuang: I think it is still a bit different, the tricky part here is that the variable `pt_idx` (i.e., the `perm`utation) contains all the information for partial transposing the basis.
Let's say I have a `QuantumObject` which contains three sub-system (`dims=[N1, N2, N3]`), and I want to partial transpose:
- the first sub-system, then `pt_idx=[4, 2, 3, 1, 5, 6]`
- the second sub-system, then `pt_idx=[1, 5, 3, 4, 2, 6]`
- or even both the first and second sub-system, then `pt_idx=[4, 5, 3, 1, 2, 6]`

I looked into the definition of `row_major_reshape`:
```
PermutedDimsArray(reshape(Q, reverse(shapes)...), (length(shapes):-1:1))
```
The argument `perm` in `PermutedDimsArray` is fixed to be `length(shapes):-1:1`, where `shapes` here equals to `(dims..., dims...)=(N1, N2, N3, N1, N2, N3)`. So I think `row_major_reshape` doesn't fit the case I need here ?

However, I didn't know `PermutedDimsArray(A, perm) -> B` in the beginning. The official documentation says that it is similar to `permutedims`, except that no copying occurs (`B` shares storage with `A`). Therefore, I think I should use `PermutedDimsArray` instead of `permutedims`.

albertomercurio: Yes, I think that it is better to use `PermutedDimsArray`

src/negativity.jl
```diff
@@ -0,0 +1,136 @@
...
+function negativity(ρ::QuantumObject, subsys::Int; logarithmic::Bool=false)
+    mask = fill(false, length(ρ.dims))
+    try
+        mask[subsys] = true
+    catch
+        error("Invalid index of subsys: $subsys")
+    end
+
+    ρ_pt = partial_transpose(ρ, mask)
+    tr_norm = sum(sqrt.(eigvals(ρ_pt' * ρ_pt)))
```
albertomercurio, line 48: Here maybe `sum(sqrt, eigvals(ρ_pt' * ρ_pt))` might be better, avoiding extra allocations.

src/negativity.jl
ytdHuang: updated in comment cd7d7a79d0fca3cbf6ef9c85877066be09f960d4

src/negativity.jl
ytdHuang: Thank you for the suggestion
I have made an update in commit cd7d7a79d0fca3cbf6ef9c85877066be09f960d4

albertomercurio (commented):
I think it is good. We can merge it. The dense case should already be agnostic. I don't know if in the future the sparse case can be optimized too.

ytdHuang:
> I think it is good. We can merge it. The dense case should already be agnostic. I don't know if in the future the sparse case can be optimized too.

Yeah, we can think a bit more about the sparse case of `partial_transpose` in the future. 

While calculating the `negativity`, it needs to solve the `eigvals` and it will always change the operator into dense matrices first. Therefore, I think the sparse case of `partial_transpose` is quite useless so far. But at least from the test, I confirmed the `partial_transpose` algorithm for sparse matrices is correct.

Maybe (if necessary) we can optimize this part after we develop methods to solve eigenvalues for sparse matrices.

albertomercurio (commented):
That' right, we almost always have dense matrices. We can merge it.

PR #33, opened 2024-03-19
Extend method for `ptrace` and `entanglement`

ytdHuang:
This is just an extend method for lazy people like me.
If we are doing partial trace and leaving only one sub-system, we can specify the `sel::Int` only (without adding the square brackets `[]` to generate a `Vector`).


albertomercurio (commented):
Great!

PR #48, opened 2024-04-01
Generalize the constructor of `QuantumObject`

### LINKED ISSUE #46: `Matrix`-type of `Ket` and `Bra`
A matrix with `size=(n, 1)` should also be considered as a `Ket`
But currently it shows an error, for example:
```julia
julia > v = Matrix([2 3])'  # this is a vector but in `Matrix`-type
julia > size(v) 
(2, 1)
julia > Qobj(v)
DimensionMismatch: matrix is not square: dimensions are (2, 1)
```

Also, I suggest to ***remove*** the following support:
```julia
julia > v = [1, 2]
julia > isbra(Qobj(v, type=BraQuantumObject))
true
```
User should do this instead:
```julia
julia > isbra(Qobj(v'))
true

ytdHuang:
fix #46 

Also add tests for expected errors in generating `QuantumObjects`.

### DIFF
--- a/src/quantum_object.jl
+++ b/src/quantum_object.jl
@@ -64,31 +64,62 @@ mutable struct QuantumObject{MT<:AbstractArray,ObjType<:QuantumObjectType} <: Ab
     dims::Vector{Int}
 end
 
-function QuantumObject(A::AbstractVector{T}; type::Type{ObjType}=KetQuantumObject, dims=nothing) where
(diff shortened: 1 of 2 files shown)


albertomercurio (approved):
Thanks! That's a better implementation for sure. I'm a bit unsure if in the future we might need tensors instead of just vectors and matrices. But for the moment this is very fine.

PR #53, opened 2024-04-04
Optimize `next!` in `ProgressBar`

ytdHuang:
Optimize the display of progress bar:

- add newline (`\n`) after the process is finished (`100.0%`)
- use floor division operator (`÷`)
- use `lpad` to make the display string length be fixed for the following variables:
  - `percentage_100` fix string length to `5` (e.g., `  5.0`, ` 32.0`, `100.0`)
  - string length of values of minute and second be fixed to `2` (e.g., `00`, `03`, `60`)

The last one makes the string length of `print` be exactly same for very short processes. Otherwise, the update makes the output shake.

### DIFF
--- a/src/progress_bar.jl
+++ b/src/progress_bar.jl
@@ -14,7 +14,6 @@ function ProgressBar(max_counts; enable=true, bar_width=30)
 end
 
 function next!(p::ProgressBar)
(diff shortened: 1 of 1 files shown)


albertomercurio (approved):
Looks great!

PR #61, opened 2024-04-07
Add `JET.jl` runtest

albertomercurio:
This feature adds `JET.jl` for type instabilities and bugs detection. For the moment it has detected several bugs, that have to be fixed in the future.


ytdHuang:
Shall we make code quality tests (including `Aqua.jl` and `JET.jl`) as an independent test `GROUP` instead of `Core` tests?

Let's say we call it `GROUP="Code Quality"`, in the end of `runtests.jl` we can add:
```julia
if (GROUP == "All") || (GROUP == "Code Quality")
    include(joinpath(testdir, "aqua.jl"))
    include(joinpath(testdir, "JET.jl"))
end
```
and remove these two tests from `core_tests` list.

In this case, the runtests CI can have an independent group which runs this `Code Quality` part of test with latest Julia version and Linux OS only. Furthermore, we can just set the `[compat]` of `Aqua` and `JET` for supporting the latest version of Julia.

albertomercurio:
Yes, good idea. I changed the files accordingly. Please check if everything is fine, I'm not an expert on CIs 😄

albertomercurio:
There are some fails on previous Julia versions. It doesn't seem to be related to this PR.

albertomercurio:
We can wait a while before they fix this issue.

albertomercurio:
> Yes, good idea. I changed the files accordingly. Please check if everything is fine, I'm not an expert on CIs 😄

For the moment, every runtest imports JET, right? Is there a way to import it only for the Code QUality CIs?

Moreover, there is an error in the current CI you just changed.

ytdHuang:
> > Yes, good idea. I changed the files accordingly. Please check if everything is fine, I'm not an expert on CIs 😄
> 
> For the moment, every runtest imports JET, right? Is there a way to import it only for the Code QUality CIs?
> 
> Moreover, there is an error in the current CI you just changed.

Yeah, I was also trying to do this.

ytdHuang:
> > Yes, good idea. I changed the files accordingly. Please check if everything is fine, I'm not an expert on CIs 😄
> 
> For the moment, every runtest imports JET, right? Is there a way to import it only for the Code QUality CIs?
> 
> Moreover, there is an error in the current CI you just changed.

I think the error comes from the `Manifest.toml`. Although we didn't import `JET` for other versions of julia, but it's still in the `[compat]` list in `Project.toml`, and thus the pre-build process throws errors.

albertomercurio:
> I think the error comes from the `Manifest.toml`. Although we didn't import `JET` for other versions of julia, but it's still in the `[compat]` list in `Project.toml`, and thus the pre-build process throws errors.

No now it is fine, you fixed the Code Quality CI error. The errors for previous Julia version doesn't seem to be related to this PR.

Anyway, I will probably make a dedicated environment for the code quality tests.

ytdHuang:
> > I think the error comes from the `Manifest.toml`. Although we didn't import `JET` for other versions of julia, but it's still in the `[compat]` list in `Project.toml`, and thus the pre-build process throws errors.
> 
> No now it is fine, you fixed the Code Quality CI error. The errors for previous Julia version doesn't seem to be related to this PR.
> 
> Anyway, I will probably make a dedicated environment for the code quality tests.

I mean, the `compat` errors between `JET.jl` and previous julia versions.
Maybe can take this [issue](https://github.com/aviatesk/JET.jl/issues/546) as a reference

albertomercurio:
There is a compilation error related to only the 1.9 version. It doesn't seem related to this PR, since the error appears even for the main branch.

Anyway, I made a separate environment for the code quality tests. What do you think?

ytdHuang (approved):

PR #72, opened 2024-04-14
Improve allocations for ProgressBar

albertomercurio:
The `ProgressBar` functionality was allocating more memory than expected. With the current version, it does not allocate memory anymore. Moreover, the `ReentrantLock` is removed, since the `print` functions are already thread-safe. I changed the `counter` as an *Atomic* object, to update it in a thread-safe manner.

### DIFF
--- a/src/progress_bar.jl
+++ b/src/progress_bar.jl
@@ -13,6 +13,8 @@ function ProgressBar(max_counts::Int; enable::Bool=true, bar_width::Int=30)
 end
 
 function next!(p::ProgressBar, io::IO=stdout)
+    p.counter[] >= p.max_counts && return
+
     Threads.atomic_add!(p.counter, 1)
 
     !p.enable && return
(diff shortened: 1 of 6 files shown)


albertomercurio (commented):
The `ProgressBar` also had the overflow error. Thus, I added a bound check, in order to not increase anymore the counter when finish.

Furthermore, I movet the Test-Quality tests before the Core tests, so that the the test results are all close to each other. In the previous configuration, after the Core tests, there was the long output for the installation of *Aqua.jl* and *JET.jl*

ytdHuang:
Nice ~ !
I was also thinking about showing progress bar even when `e_ops` is empty.

src/progress_bar.jl
```diff
@@ -36,17 +34,16 @@ function next!(p::ProgressBar, io::IO=stdout)
     elapsed_time_str = string(elapsed_time ÷ 3600, "h ", lpad((elapsed_time % 3600) ÷ 60, 2, "0"), "m ", lpad(elapsed_time % 60, 2, "0"), "s")
 
     # Calculate the estimated time of arrival
-    eta = floor(Int, elapsed_time ÷ counter * (max_counts - counter))
+    eta = floor(Int, elapsed_time / counter * (max_counts - counter))
```
ytdHuang, line 37: just a small comment, would it be better to change this line as
```julia
eta = elapsed_time * (max_counts - counter) ÷ counter
```
since `elapsed_time`, `max_counts`, `counter` are all `integers`
by using floor division (`÷`), we don't need to do `floor()` again after normal division (`/`) ?

ytdHuang (approved):

src/progress_bar.jl
albertomercurio: The rpevious case had some issues, because it showed ETA = 0s while it was supposed to be 40s ish. So I thought to fix this problem in this way

PR #73, opened 2024-04-14
Improve WignerClenshaw allocations

albertomercurio:
The previous method had a lot of memory allocations. With this fix, I made most of the operations inplace.


albertomercurio:
I don't know why it failed in the latest macOS. Should we leave it as it is?

ytdHuang:
> I don't know why it failed in the latest macOS. Should we leave it as it is?

I think that error is something we cannot handle it.
Someone will fix it for us lol

ytdHuang (approved):

PR #78, opened 2024-04-23
Element type conversion for `QuantumObject`

ytdHuang:
Support the following type conversion:
```julia
using QuantumToolbox
v = Qobj(Int64[1, 1])
m = Qobj(Int64[1, 1])

Vector{ComplexF64}(v)
Matrix{ComplexF64}(m)
```
and so does `SparseVector` and `SparseMatrixCSC`


albertomercurio (approved):
Great feature!

albertomercurio:
I would say also to add the `sparse` command. I don't remember that this is already implemented, right?

albertomercurio:
Mmmm ok, there is already `dense_to_sparse`. We can merge it as it is

PR #93, opened 2024-05-01
Fix local `JET` runtest

ytdHuang:
For local development, only run code quality tests under `VERSION >= v"1.9"`
Because `JET.test_package` does not support for `Julia v1.7`, and some errors will occur under `Julia v1.8`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -24,7 +24,7 @@ core_tests = [
     "wigner.jl",
 ]
 
-if (GROUP == "All") || (GROUP == "Code-Quality")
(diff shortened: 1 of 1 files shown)


albertomercurio:
Why does the documentation fail?

ytdHuang:
> Why does the documentation fail?

The building process somehow breaks at precompilation of the dependencies. I think there are some issues between `Plots.jl` and `GR_jll`.
Someone might fix it in the future.

albertomercurio (approved):

albertomercurio:
Ok, great.

PR #96, opened 2024-05-04
Add benchmark tracking

albertomercurio:
This PR introduces the feature of tracking the benchmark results over the commits, and produces warnings on PRs if the benchmarks regress.

It uses [https://github.com/benchmark-action/github-action-benchmark/](https://github.com/benchmark-action/github-action-benchmark/) and I used their examples as template, together with the [Lux.jl](https://github.com/LuxDL/Lux.jl) use.

Currently, I have added only the **steadystate** case, just to make it very minimal. Other benchmarks could be introduced in the following PRs, after checking that this actions works fine.

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -1,0 +1,60 @@
+name: Benchmark Tracking
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+permissions:
+    contents: write
+    deployments: write
+
+concurrency:
+    # Skip intermediate builds: always.
+    # Cancel intermediate builds: only if it is a pull request build.
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: '1'
+          arch: x64
+      - uses: actions/cache@v4
+        env:
+          cache-name: cache-artifacts
+        with:
+          path: ~/.julia/artifacts
+          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-test-${{ env.cache-name }}-
+            ${{ runner.os }}-test-
+            ${{ runner.os }}-
+      - name: Run benchmark
+        run: |
+          cd benchmarks
+          julia --project --threads=2 --color=yes -e '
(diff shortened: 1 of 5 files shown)


benchmarks/.gitignore
```diff
@@ -0,0 +1 @@
+/benchmarks_output.json
```
ytdHuang, line 1: should we just put a line in the original `.gitignore` file, like `*.json` (just ignore all `json` files)
I think it would be better to gather all the `.gitignore files together`

benchmarks/runbenchmarks.jl
```diff
@@ -0,0 +1,28 @@
+using BenchmarkTools
+using QuantumToolbox
+
+suite = BenchmarkGroup()
+
+suite["steadystate"] = BenchmarkGroup(["steadystate"])
+
+
+## steadystate ##
```
ytdHuang, line 9: There might be many benchmarks in the future. Should we separate each benchmark topic into individual files and `include` them in this file? 

To avoid messing the in-scope variables, maybe can put each benchmark into a function and just return `@benchmarkable XXX()`
and call those functions in this file:
```julia
suite["XXX"]["YYY"] = f(...)
```

What do you think ?

ytdHuang:
I think I will also modify the CI pipeline config file later.

benchmarks/.gitignore
albertomercurio: Ok fine, although it's too restrictive maybe. But for the moment we are not using the json files, so it's fine.

benchmarks/runbenchmarks.jl
albertomercurio: Yes I agree, this is just to have everything minimal for the moment, and see if everything work. Then we should add as many benchmarks as we want, each one in a separated file.

.github/workflows/Benchmarks.yml
```diff
@@ -20,22 +30,14 @@ concurrency:
...
   benchmark:
     runs-on: ubuntu-latest
+    if: ${{ !github.event.pull_request.draft }}
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
           version: '1'
           arch: x64
-      - uses: actions/cache@v4
```
albertomercurio: Why don't we need these lines? I copied them from both the **Lux.jl** and the original example of the action

ytdHuang: I replace these lines with:
```
- uses: julia-actions/cache@v1
```

benchmarks/runbenchmarks.jl
ytdHuang: Okay~

.github/workflows/Benchmarks.yml
albertomercurio: Ok I see!

ytdHuang (approved):

PR #98, opened 2024-05-05
Add more benchmarks

albertomercurio:
Add more benchmarks for benchmark tracking. Add also the menu link for the Bechmarks page


albertomercurio:
The Julia cache gives some errors. Using back again the cache management used in the example of the action resolved this issue. Already in the previous PR #97 there was this error, but I had directly merged since I changed  only the readme file.

PR #101, opened 2024-05-18
Several minor changes to eigsolve

albertomercurio:
This improves a bit the current `eigsolve` algorithm. Now it can detect covergence just after the first iteration. The code is also much more clean.

I have also changed the tolerances for the LinearSolve, in case of shift-inverse, since I faced instabilities in delicate cases.


ytdHuang:
@albertomercurio 
Does the final minor changes allow us to drop any import functions in [src/QuantumToolbox.jl](https://github.com/albertomercurio/QuantumToolbox.jl/blob/main/src/QuantumToolbox.jl#L15-L22) ?

ytdHuang (approved):

albertomercurio:
No, it’s only more clean and efficient

PR #102, opened 2024-05-18
Support `svdvals`, generalize `norm`, and introduce `tracedist` for `Qobj`

ytdHuang:
The original `LinearAlgebra.norm` calculates standard vector norm no matter the input array is vector or matrix. But usually we calculate [Schatten p-norm](https://en.wikipedia.org/wiki/Schatten_norm) for matrices (density operators), which need to first calculate the singular values before taking the standard vector norm.

Summary for this PR:

- Support `LinearAlgebra.svdvals` (singular values) for `Qobj`
- Generalize `LinearAlgebra.norm(A::QuantumObject, p::Real=2)`:
  - if `Qobj` is `Ket`, `Bra`, `OperatorKet`, or `OperatorBra`, calculate the standard vector p-norm
  - if `Qobj` is `Operator` or `SuperOperator`, calculate the [Schatten p-norm](https://en.wikipedia.org/wiki/Schatten_norm)
- Since the definition of `negativity` is related to the trace-norm (Schatten 1-norm), use the above method directly can slightly increase the efficiency.
- Introduce `tracedist` (also relates to the trace-norm) between two given `Qobj`

### DIFF
--- a/src/general_functions.jl
+++ b/src/general_functions.jl
@@ -1,6 +1,6 @@
 export get_data, get_coherence, expect, ptrace
 export mat2vec, vec2mat
-export entropy_vn, entanglement
+export entropy_vn, entanglement, tracedist
 export gaussian, n_th
 
 export row_major_reshape, tidyup, tidyup!, meshgrid, sparse_to_dense, dense_to_sparse
@@ -328,9 +328,18 @@ function n_th(ω::Real, T::Real)::Float64
     return 1 / (exp(ω / T) - 1)
 end
 
+@doc raw"""
+    tracedist(A::QuantumObject, B::QuantumObject)
 
+Calculates the [trace distance](https://en.wikipedia.org/wiki/Trace_distance) between two [`QuantumObject`](@ref) `A` and `B`:
+``T(A, B) = frac{1}{2} \lVert A - B \rVert_1``
 
-
+Note that `A` and `B` must be either [`Ket`](@ref) or [`Operator`](@ref).
+"""
+tracedist(A::QuantumObject{<:AbstractArray{T1},OperatorQuantumObject}, B::QuantumObject{<:AbstractArray{T2},OperatorQuantumObject}) where {T1,T2} = 0.5 * norm(A - B, 1)
+tracedist(A::QuantumObject{<:AbstractArray{T1},KetQuantumObject}, B::QuantumObject{<:AbstractArray{T2},OperatorQuantumObject}) where {T1,T2} = tracedist(ket2dm(A), B)
+tracedist(A::QuantumObject{<:AbstractArray{T1},OperatorQuantumObject}, B::QuantumObject{<:AbstractArray{T2},KetQuantumObject}) where {T1,T2} = tracedist(A, ket2dm(B))
+tracedist(A::QuantumObject{<:AbstractArray{T1},KetQuantumObject}, B::QuantumObject{<:AbstractArray{T2},KetQuantumObject}) where {T1,T2} = tracedist(ket2dm(A), ket2dm(B))
 
 function _ptrace_ket(QO::AbstractArray{T1}, dims::Vector{<:Integer}, sel::Vector{T2}) where
     {T1,T2<:Integer}
--- a/src/general_functions.jl
+++ b/src/general_functions.jl
@@ -329,17 +329,14 @@ function n_th(ω::Real, T::Real)::Float64
 end
 
 @doc raw"""
-    tracedist(A::QuantumObject, B::QuantumObject)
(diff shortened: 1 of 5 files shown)


src/general_functions.jl
```diff
@@ -328,9 +328,18 @@ function n_th(ω::Real, T::Real)::Float64
...
+@doc raw"""
+    tracedist(A::QuantumObject, B::QuantumObject)
 
+Calculates the [trace distance](https://en.wikipedia.org/wiki/Trace_distance) between two [`QuantumObject`](@ref) `A` and `B`:
+``T(A, B) = frac{1}{2} \lVert A - B \rVert_1``
 
-
+Note that `A` and `B` must be either [`Ket`](@ref) or [`Operator`](@ref).
+"""
+tracedist(A::QuantumObject{<:AbstractArray{T1},OperatorQuantumObject}, B::QuantumObject{<:AbstractArray{T2},OperatorQuantumObject}) where {T1,T2} = 0.5 * norm(A - B, 1)
```
albertomercurio, line 339: Should we avoid the 0.5 and directly dividing by 2? In this way we avoid to promote it to Float64, if not. Am I wrong?

src/general_functions.jl
```diff
@@ -328,9 +328,18 @@ function n_th(ω::Real, T::Real)::Float64
...
+tracedist(A::QuantumObject{<:AbstractArray{T1},KetQuantumObject}, B::QuantumObject{<:AbstractArray{T2},OperatorQuantumObject}) where {T1,T2} = tracedist(ket2dm(A), B)
+tracedist(A::QuantumObject{<:AbstractArray{T1},OperatorQuantumObject}, B::QuantumObject{<:AbstractArray{T2},KetQuantumObject}) where {T1,T2} = tracedist(A, ket2dm(B))
+tracedist(A::QuantumObject{<:AbstractArray{T1},KetQuantumObject}, B::QuantumObject{<:AbstractArray{T2},KetQuantumObject}) where {T1,T2} = tracedist(ket2dm(A), ket2dm(B))
```
albertomercurio, line 342: I was wondering if it is the case to define the case ‘ket2dm(Operator) = Operator’. In this way we avoid the two cases above. What do you think?

src/negativity.jl
```diff
@@ -89,7 +89,7 @@ function _partial_transpose(ρ::QuantumObject{<:AbstractArray, OperatorQuantumOb
         [pt_dims[n, 3 - mask2[n]] for n in 1:nsys]  # opposite value in mask2 (1 -> 2, and 2 -> 1)
     ]
     return QuantumObject(
-        reshape(PermutedDimsArray(reshape(ρ.data, (ρ.dims..., ρ.dims...)), pt_idx), size(ρ)),
+        Matrix(reshape(PermutedDimsArray(reshape(ρ.data, (ρ.dims..., ρ.dims...)), pt_idx), size(ρ))),
```
albertomercurio, line 92: I don’t remember if this function is gpu compatible. If yes, should ve convert it into the corresponding type?

src/general_functions.jl
ytdHuang: But usually the output of `tracedist` is `Float`

src/negativity.jl
ytdHuang: This is pretty wierd.
if I use: `reshape(PermutedDimsArray(reshape(ρ.data, (ρ.dims..., ρ.dims...)), pt_idx), size(ρ))`, it returns something in the type of:
```
Base.ReshapedArray{Float64, 2, PermutedDimsArray{Float64, 4, (1, 4, 3, 2), (1, 4, 3, 2), Array{Float64, 4}}, Tuple{Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}, Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}, Base.MultiplicativeInverses.SignedMultiplicativeInverse{Int64}}}
```

But if I use: `reshape(permutedims(reshape(ρ.data, (ρ.dims..., ρ.dims...)), pt_idx), size(ρ))`, it just returns in the type of `Matrix{Float64}`.

This causes type-instability, maybe we should just use `permutedims` (as I updated in ) in this case, or do you have other ideas that would work?

src/general_functions.jl
ytdHuang: I see, maybe we just change it here. (updated in https://github.com/albertomercurio/QuantumToolbox.jl/pull/102/commits/ef01d3b489b625dc470802dde1125c075ec6c788)

src/negativity.jl
albertomercurio: Ok

albertomercurio (approved):

PR #106, opened 2024-05-20
Adjust Logo size and position in README

ytdHuang:


albertomercurio:
I added a dot in the beginning of the log path, because my local preview was failing to find the logo png

albertomercurio (approved):

ytdHuang:
Seems like there are too many badges
Let me change the order of the badges?

PR #111, opened 2024-05-20
Format Documents

albertomercurio:
I executed the JuliaFormatter.jl to format all the source code files.

I added also the [GitHub Action](https://github.com/julia-actions/julia-format), which creates PRs to correct the format.


.JuliaFormatter.toml
```diff
@@ -0,0 +1,6 @@
+always_for_in=true
+remove_extra_newlines=true
+long_to_short_function_def=true
+format_docstrings=true
+indent_submodule=true
+format_markdown=true
```
ytdHuang, line 6: maybe add the following format settings ?
- [`whitespace_ops_in_indices=true`](https://domluna.github.io/JuliaFormatter.jl/dev/#whitespace_ops_in_indices)
- [`whitespace_typedefs=true`](https://domluna.github.io/JuliaFormatter.jl/dev/#whitespace_typedefs)

ext/QuantumToolboxCUDAExt.jl
```diff
@@ -10,56 +10,64 @@ import SparseArrays: SparseVector, SparseMatrixCSC
...
+    QuantumObject(CuArray(A.data), A.type, A.dims)
 
 @doc raw"""
     CuArray{T}(A::QuantumObject)
 
 If `A.data` is a dense array, return a new [`QuantumObject`](@ref) where `A.data` is in the type of `CUDA.CuArray` with element type `T` for gpu calculations.
 """
-CuArray{T}(A::QuantumObject{Tq}) where {T,Tq<:Union{Vector,Matrix}} = QuantumObject(CuArray{T}(A.data), A.type, A.dims)
+CuArray{T}(A::QuantumObject{Tq}) where {T,Tq<:Union{Vector,Matrix}} =
+    QuantumObject(CuArray{T}(A.data), A.type, A.dims)
```
ytdHuang, line 22: Should we increase the `margin=120` (`default:92`)?
Since our type names (e.g., `QuantumObject{<:AbstractArray{T}}`, `SuperOperatorQuantumObject`, and so on) is quite long. Many single line functions will become two lines.

The case I tagged here has (original) `length=113`.

.github/workflows/Format.yml
```diff
@@ -0,0 +1,31 @@
+name: format-pr
```
ytdHuang, line 1: Most of our CI pipeline names starts with capital letters, maybe change it as `Format-PR` or `JuliaFormatter-PR` ?

ytdHuang: If we are using `JuliaFormatter-PR` as the name, maybe also change the filename as `JuliaFormatter.yml` ?

.JuliaFormatter.toml
albertomercurio: I don’t know if I like the white spaces. Regarding the return, yes, better to make it always true

ytdHuang: okay~

ytdHuang (approved):

PR #112, opened 2024-05-21
Develop documentation

ytdHuang:
This is the PR for developing the documentation.

The structure of the pages will be organized in the `const` variable called `PAGES` in `docs/make.jl`.

Currently, the documentation will be divided into five parts:
- Front Matter: including package introduction, installation
- Users Guide: follow the similar content in [QuTiP Users Guide](https://qutip.org/documentation.html)
- Tutorials
- Solver Benchmarks
- API


README.md
```diff
@@ -25,14 +25,88 @@ and [Y.-T. Huang](https://github.com/ytdHuang).
...
-- **Measurement and Statistics:** Comprehensive quantum measurement simulation and analysis.
-- **GPU and Distributed Computing:** Leverage GPU and distributed resources for high-performance computing.
+- **Quantum State and Operator Manipulation:** Easily handle quantum states and operators with a rich set of tools, with the same functionalities as QuTiP.
+- **Dynamical Evolution:** Advanced solvers for time evolution of quantum systems, thanks to the powerful [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl) package.
+- **GPU Computing:** Leverage GPU resources for high-performance computing. For example, you run the master equation direclty on the GPU with the same syntax as the CPU case.
+- **Distributed Computing:** Distribute the computation over multiple nodes (e.g., a cluster). For example, you can run undreds of quantum trajectories in parallel on a cluster, with, again, the same syntax as the simple case.
+- **Easy Extension:** Easily extend the package, taking advantage of the Julia language features, like multiple dispatch and metaprogramming.
+
+## Brief example
```
ytdHuang, line 40: How about moving the entire `## Brief Example` after the `## Installation` ?
I think it is more nature for the users to follow: install and then how to use it.

docs/src/index.md
```diff
@@ -2,9 +2,89 @@
...
+
+QuantumToolbox.jl is equipped with a robust set of features:
+
+- **Quantum State and Operator Manipulation:** Easily handle quantum states and operators with a rich set of tools, with the same functionalities as QuTiP.
+- **Dynamical Evolution:** Advanced solvers for time evolution of quantum systems, thanks to the powerful [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl) package.
+- **GPU Computing:** Leverage GPU resources for high-performance computing. For example, you run the master equation direclty on the GPU with the same syntax as the CPU case.
+- **Distributed Computing:** Distribute the computation over multiple nodes (e.g., a cluster). For example, you can run undreds of quantum trajectories in parallel on a cluster, with, again, the same syntax as the simple case.
+- **Easy Extension:** Easily extend the package, taking advantage of the Julia language features, like multiple dispatch and metaprogramming.
+
+## Brief example
```
ytdHuang, line 21: Maybe make this `Brief example` as a new page in `Front Matter` (after `Installation` or `Key differences from QuTiP`) and title as `Quick Start` ?
Cause I think it is better to follow the installation first and then know how to use the code.
What do you think?

docs/src/index.md
```diff
@@ -2,9 +2,89 @@
...
-# Introduction
+# [Introduction](@id doc: Introduction)
 
-[QuantumToolbox](https://github.com/qutip/QuantumToolbox.jl) is a Julia package for numerical simulation of quantum systems.
+[QuantumToolbox.jl](https://github.com/qutip/QuantumToolbox.jl) is a cutting-edge Julia package designed for quantum physics simulations, closely emulating the popular Python [QuTiP](https://github.com/qutip/qutip) package. It uniquely combines the simplicity and power of Julia with advanced features like GPU acceleration and distributed computing, making simulation of quantum systems more accessible and efficient.
 
-```@index
+*With this package, moving from Python to Julia for quantum physics simulations has never been easier*, due to the similar syntax and functionalities.
+
+## Features
```
ytdHuang, line 11: The sub-title (`##`) cause `Introduction` to be extendable in the sidebar of the documentation.
Maybe make it as sub-sub-title (`###`) ?

docs/src/install.md
```diff
@@ -0,0 +1,23 @@
+# [Installation](@id doc: Installation)
+
+## QuantumToolbox.jl
```
ytdHuang, line 3: Maybe also make this as sub-sub-title (`###`) to avoid unnecessary expansion in sidebar.

docs/src/install.md
```diff
@@ -0,0 +1,23 @@
...
+```
+Alternatively, this can also be done in Julia's [Pkg REPL](https://julialang.github.io/Pkg.jl/v1/getting-started/) by pressing the key `]` in the REPL to use the package mode, and then type the following command:
+```julia-REPL
+(1.7) pkg> add QuantumToolbox
+```
+More information about `Julia`'s package manager can be found at [`Pkg.jl`](https://julialang.github.io/Pkg.jl/v1/).  
+!!! note "Julia 1.7"
+    `QuantumToolbox.jl` requires Julia 1.7 or higher.
+
+## Verify the installation
```
ytdHuang, line 17: Maybe also make this as sub-sub-title (`###`) to avoid unnecessary expansion in sidebar

docs/src/install.md
```diff
@@ -0,0 +1,23 @@
...
+To load the package and check the version information, use either `versioninfo()` or `about()`, namely
+```julia
+julia> using QuantumToolbox
+julia> QuantumToolbox.versioninfo()
+julia> QuantumToolbox.about()
```
ytdHuang, line 22: When you address the above comments, could you also remove the three `julia> ` here ?

docs/make.jl
```diff
@@ -3,11 +3,57 @@ using Documenter
...
+            "users_guide/states_and_operators/qubit_systems.md",
+            "users_guide/states_and_operators/expectation_values.md",
+            "users_guide/states_and_operators/superoperators.md",
+        ],
+        "Tensor Products and Partial Traces" => [
+            "users_guide/tensor_product/tensor.md",
+            "users_guide/tensor_product/partial_trace.md",
+        ],
+        "Time Evolution and Dynamics" => [
+            "users_guide/time_evolution/intro.md"
```
ytdHuang, line 30: Maybe make the label in left-sidebar as `Introduction` ?
```
"Time Evolution and Dynamics" => [
    "Introduction" => "users_guide/time_evolution/intro.md",
],
```

ytdHuang:
Ahh,
Maybe next time we can consider to make the change of `README` as an independent PR. So that the CI pipeline will skip the runtests automatically.

ytdHuang:
The JuliaFormatter is raising error in markdown files again.
But I can still build the documentation locally.

Should we just modify the .JuliaFormatter.toml as:
```
margin=120
always_for_in=true
remove_extra_newlines=true
long_to_short_function_def=true
always_use_return=true
format_docstrings=true
indent_submodule=true
format_markdown=false
ignore = []
```
change `format_markdown` to `false` and clear the ignore list.
P.S. I keep the blank ignore list just in case that we will ignore some files in the future. But you can remove it if you don't like.

docs/src/index.md
albertomercurio: I don’t know if moving to a new page is good. For sure we can change the order and put the installation instruction first. Because I think that it is important that the reader knows the simplicity of moving from QuTiP, already in the first page.

docs/src/index.md
albertomercurio: In my opinion, Inlike this. Other documentations use this, from Julia itself, Documenter and others.

docs/src/install.md
albertomercurio: I think that we can move the subtitle `## Introdution`` under the introduction, before the first example. Such that we ha a single page, with every information inside. Indeed, most of the people will read only the first page, and so this page is very important

docs/src/install.md
albertomercurio: Yes, this one is better to make as a sub-sub-title

docs/src/install.md
albertomercurio: Ok

docs/src/index.md
ytdHuang: Let us discuss this in `zulip`?

docs/src/install.md
ytdHuang: Let us discuss this in `zulip`?

README.md
ytdHuang: Let us discuss this in `zulip`?

albertomercurio (approved):

PR #118, opened 2024-05-22
Introduce CI FormatCheck

ytdHuang:
In this PR, I replace the original Auto-JuliaFormatter-PR CI as Format-Check.

This is just a simple check whether the code is formatted before merging.


albertomercurio:
Why don’t we use the proposed code of the action webpage?

```yaml
name: Format suggestions
on:
  pull_request:
    # this argument is not required if you don't use the `suggestion-label` input
    types: [ opened, reopened, synchronize, labeled, unlabeled ]
jobs:
  code-style:
    runs-on: ubuntu-latest
    steps:
      - uses: julia-actions/julia-format@v3
        with:
          version: '1' # Set `version` to '1.0.54' if you need to use JuliaFormatter.jl v1.0.54 (default: '1')
          suggestion-label: 'format-suggest' # leave
```

BTW, we should try to avoid to rename CIs, otherwise we accumulate all the names in the Actions page

ytdHuang:
To avoid conflicts, I will push the formatted codes after PR #117 is merged.

ytdHuang:
> Why don’t we use the proposed code of the action webpage?
> 
> ```yaml
> name: Format suggestions
> on:
>   pull_request:
>     # this argument is not required if you don't use the `suggestion-label` input
>     types: [ opened, reopened, synchronize, labeled, unlabeled ]
> jobs:
>   code-style:
>     runs-on: ubuntu-latest
>     steps:
>       - uses: julia-actions/julia-format@v3
>         with:
>           version: '1' # Set `version` to '1.0.54' if you need to use JuliaFormatter.jl v1.0.54 (default: '1')
>           suggestion-label: 'format-suggest' # leave
> ```
> 
> BTW, we should try to avoid to rename CIs, otherwise we accumulate all the names in the Actions page

They mentioned some issues in their [README](https://github.com/julia-actions/julia-format?tab=readme-ov-file). I think it would be better for us to just write it by ourselves.

ytdHuang:
I have disabled the outdated CI workflows in action page

.github/workflows/FormatCheck.yml
```diff
@@ -0,0 +1,44 @@
...
+      - synchronize
+      - ready_for_review
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: ${{ !github.event.pull_request.draft }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
```
albertomercurio, line 23: Can we use the cache action, or not?

albertomercurio (approved):

.github/workflows/FormatCheck.yml
albertomercurio: Otherwise we can merge.

ytdHuang: not sure whether need the cache, since this CI only need to install `JuliaFormatter.jl`

ytdHuang:
The JuliaFormatter keep formatting `README.md` into a wrong one.

albertomercurio:
The format check is failing. Is that ok?

albertomercurio:
Mmmm that’s strange

ytdHuang:
> The format check is failing. Is that ok?

The only file that is not formatted now is `README`, but `JuliaFormatter` will format it incorrectly.

ytdHuang:
> Ok, so let’s merge?

I add an `ignore = ["README.md"]` to `.JuliaFormatter.toml`.

albertomercurio:
Ok we can merge it

PR #132, opened 2024-05-26
Introduce `sqrtm` and `fidelity`

ytdHuang:
Summary of this PR:
- improve `LinearAlgebra.sqrt` to support `SparseArrays` (by using `sparse_to_dense` before `sqrt`)
- introduce `sqrtm` (same as the one in `qutip`, only works for `Operator`)
- introduce `fidelity`

### DIFF
--- a/src/general_functions.jl
+++ b/src/general_functions.jl
@@ -1,6 +1,6 @@
 export get_data, get_coherence, expect, ptrace
 export mat2vec, vec2mat
-export entropy_vn, entanglement, tracedist
+export entropy_vn, entanglement, tracedist, fidelity
 export gaussian, n_th
 
 export row_major_reshape, tidyup, tidyup!, meshgrid, sparse_to_dense, dense_to_sparse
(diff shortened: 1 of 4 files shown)


src/general_functions.jl
```diff
@@ -311,6 +311,37 @@ function expect(
...
+    ψ::QuantumObject{<:AbstractArray{T2},KetQuantumObject},
+) where {T1,T2} = sqrt(abs(expect(ρ, ψ)))
+fidelity(
+    ψ::QuantumObject{<:AbstractArray{T1},KetQuantumObject},
+    σ::QuantumObject{<:AbstractArray{T2},OperatorQuantumObject},
+) where {T1,T2} = fidelity(σ, ψ)
+fidelity(
+    ψ::QuantumObject{<:AbstractArray{T1},KetQuantumObject},
+    ϕ::QuantumObject{<:AbstractArray{T2},KetQuantumObject},
+) where {T1,T2} = abs(ψ' * ϕ)
```
albertomercurio, line 343: Should we perform `abs(dot(a, b))` instead?

src/quantum_object.jl
```diff
@@ -741,7 +742,17 @@ LinearAlgebra.rmul!(B::QuantumObject{<:AbstractArray}, a::Number) = (rmul!(B.dat
 @inline LinearAlgebra.mul!(y::AbstractVector{Ty}, A::QuantumObject{<:AbstractMatrix{Ta}}, x, α, β) where {Ty,Ta} =
     mul!(y, A.data, x, α, β)
 
-LinearAlgebra.sqrt(A::QuantumObject{<:AbstractArray{T}}) where {T} = QuantumObject(sqrt(A.data), A.type, A.dims)
+LinearAlgebra.sqrt(A::QuantumObject{<:AbstractArray{T}}) where {T} =
+    QuantumObject(sqrt(sparse_to_dense(A.data)), A.type, A.dims)
```
albertomercurio, line 746: We need `sparse_to_dense` because `sqrt` is not supported for sparse?

src/general_functions.jl
ytdHuang: sure~

src/quantum_object.jl
ytdHuang: Yes, I tried applying it on sparse matrices directly. But because sqrt uses eigen, which is not supported for sparse matrices.

albertomercurio (approved):

PR #133, opened 2024-05-26
Update License, Code of Conduct, and README

ytdHuang:
Summary of this PR:
- Update to same license as `qutip` and add license badge in `README`
- Add code of conduct (same as `qutip`) and also ignore it in `JuliaFormatter` (since it is also `markdown` file)
- Add the name list of admin team (and corresponding hyperlinks to their github page)

I find out that when someone wants to contribute (to this package) and make a PR, the list of reviewers includes all the qutip admins. So I thought we can at least do the same thing as `qutip`: put the admin name list on top of the README.

@albertomercurio, @lgravina1997
What do you think? Is it okay for you guys ?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,5 @@
+BSD 3-Clause License
+
 Copyright (c) 2022 to 2024 inclusive, QuTiP developers and contributors.
 All rights reserved.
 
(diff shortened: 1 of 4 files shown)


LICENSE
```diff
@@ -1,21 +1,27 @@
-MIT License
+Copyright (c) 2022 to 2024 inclusive, QuTiP developers and contributors.
```
albertomercurio, line 1: I would suggest to add the following text as first line, like the orginsl github template for this license

```
BSD 3-Clause License
```

ytdHuang: Okay~ Done

albertomercurio (approved):

Issue #81, opened 2024-04-28
Add a the iterative solvers and eigensolver for `steadystate`


albertomercurio:
There is only one [`steadystate`](https://github.com/qutip/QuantumToolbox.jl/blob/a227f5f088e555d25810f7ab519d2b860d5f8735/src/steadystate.jl#L15) solver: the direct solver.

It could be very useful to add other solvers, such as all the iterative solvers that are available through the [**LinearSolve.jl**](https://github.com/SciML/LinearSolve.jl) package and the internal [*eigensolver*](https://github.com/qutip/QuantumToolbox.jl/blob/a227f5f088e555d25810f7ab519d2b860d5f8735/src/eigsolve.jl#L367).

The eigensolver uses the internal sparse solver, that uses the Krylov subspace, taking only the eigenstate with zero eigenvalue.

Perhaps, one can also add the support for a cache (it could be very easy using the **LinearSolve.jl** package).

golanor:
Hi,

I'm trying to understand how to approach this. Is the idea to perform different types of eigendecompositions of the Lindblandian, and then return the eigenstate with zero eigenvalue?

albertomercurio:
Hi,

the idea is to add other solvers to the `steadystate` function. We might have two different main methods:
- A linear solver method (using **LinearSolve.jl** perhaps)
- An eigensolver method

The **LinearSolve.jl** has different methods inside, and so we can choose several sub methods from it. For example, this package is able to solve a linear system using both the direct solver or the iterative solvers.

Regarding the eigensolver, yes, you can solve the Liouvillian eigenvalues and then take the eigenstate with zero eigenvalue. You can use the internal `eigsolve` function, that is very efficient for solving only one eigenvalue.

LorenzoFioroni:
This is a comment to allow the Unitary Hack bots to find the PR above ( #149)

PR #149, opened 2024-05-29
Implement steady state solver using LinearSolve.jl and eigen

LorenzoFioroni:
The goal of this pull request is to make a step forward in the direction of fixing issue #81

## General description

In issue #81 it has been asked to implement, along with the `SteadyStateDirectSolver` already present within the library, an iterative solver using LinearSolve.jl and an eigen solver using the already implemented `eigsolve` function.

I aimed to make the code as agnostic as possible and compatible with both dense and sparse matrices on CPU and GPU. I haven't been able to make some combinations of method-device work yet, but I think this might be a good starting point.

## Summary of the changes

I introduced, along with the already implemented `SteadyStateDirectSolver`, two new structs: `SteadyStateEigenSolver` and `SteadyStateLinearSolver`. While the first two don't require any argument when created, the latter allows specifying optional parameters (see below).

### Direct solver

No changes have been made on this side, and the code remains as originally implemented in the library. 

From a few tests, this method seems to work on sparse and dense CPU matrices and dense GPU ones, but not on sparse GPU matrices. It seems that the left-division operation on GPU matrices results in an indexing of the array, which is disabled by `CUDA.allowscalar(false)` for efficiency reasons.

### Linear solver

The `SteadyStateLinearSolver` accepts three optional keyword arguments, allowing the specification of the solving algorithm from the LinearSolve.jl package, and the left- and right- preconditioners. Since the preconditioners are applied to a matrix `L_tmp` analogous to that defined within the direct solver (see [here](https://github.com/qutip/QuantumToolbox.jl/blob/9786ddf796fa29a23766a5cba0ed6397ad34ead7/src/steadystate.jl#L45)), they are defined as functions. For example, `SteadyStateLinearSolver(alg=KrylovJL_GMRES(), Pl=L_tmp -> ilu(L_tmp, τ = 0.01))` creates a `SteadyStateLinearSolver` which uses the `KrylovJL_GMRES` algorithm defined in LinearSolve.jl and performs a preconditioning with the `ilu` method from the IncompleteLU.jl package. If no algorithm is specified, LinearSolve.jl uses a direct method analogous to our DirectSolver. Moreover, additional parameters can be given to the `solve` function of LinearSolve.jl by passing them as keyword arguments to the call to `steadystate`.

When used as a direct solver, it seems that this method only fails for sparse GPU arrays. I think the reason might be the call to `SparseMatrixCSC()` in [LinearSolve.jl](https://github.com/SciML/LinearSolve.jl/blob/270b56d864e91c2873d27c71b78115a04b824e23/src/factorization.jl#L1359C1-L1360C1), which is not implemented when the parameters are GPU arrays

When used with an iterative method (e.g., `KrylovJL_GMRES`), this solver works for all the combinations of sparse-dense matrices and both on CPU and GPU. Yet, the application to sparse GPU matrices is very slow compared to that on CPU, because no preconditioner is applied. From the [preconditioners page](https://docs.sciml.ai/LinearSolve/stable/basics/Preconditioners/) of LinearSolve.jl's docs, the only GPU-ready preconditioners are defined in KrylovPreconditioners.jl. However, taking for example the `kp_ilu0` method, this seems to be only defined for real CSC matrices ([see here](https://github.com/JuliaSmoothOptimizers/KrylovPreconditioners.jl/blob/09af6873d27e425f8a757e4c08ae5faa0fd9b664/ext/CUDA/ilu0.jl#L97)) while it would be implemented for complex ones as well if they were CSR ([here](https://github.com/JuliaSmoothOptimizers/KrylovPreconditioners.jl/blob/09af6873d27e425f8a757e4c08ae5faa0fd9b664/ext/CUDA/ilu0.jl#L83)).

### EigenSolver

The EigenSolver is nothing but an interface to the already implemented `eigsolve` function. Any keyword arguments specified in the call to the `steadystate` function will be passed to `eigsolve`.

This method seems to work only for CPU matrices, as I consistently get a _scalar indexing_ error from the `arnoldi_init` function for GPU matrices.

---

I hope this can be a good starting point to improve the support of different solving methods within the `steadystate` function. Please let me know what you think and if there is anything I could improve already. Thank you!

### DIFF
--- a/src/steadystate.jl
+++ b/src/steadystate.jl
@@ -5,10 +5,11 @@ abstract type SteadyStateSolver end
 
 struct SteadyStateDirectSolver <: SteadyStateSolver end
 struct SteadyStateEigenSolver <: SteadyStateSolver end
-Base.@kwdef struct SteadyStateLinearSolver{MT<:Union{LinearSolve.SciMLLinearSolveAlgorithm, Nothing}} <: SteadyStateSolver 
+Base.@kwdef struct SteadyStateLinearSolver{MT<:Union{LinearSolve.SciMLLinearSolveAlgorithm,Nothing}} <:
+                   SteadyStateSolver
     alg::MT = nothing
-    Pl::Union{Function, Nothing} = nothing
-    Pr::Union{Function, Nothing} = nothing
+    Pl::Union{Function,Nothing} = nothing
+    Pr::Union{Function,Nothing} = nothing
 end
 
 function steadystate(
@@ -52,15 +53,14 @@ function _steadystate(
     fill!(datas, weight)
     Tn = sparse(rows, cols, datas, N^2, N^2)
     L_tmp = L_tmp + Tn
-    
+
     (haskey(kwargs, :Pl) || haskey(kwargs, :Pr)) && error("The use of preconditioners must be defined in the solver.")
     if !isnothing(solver.Pl)
         kwargs = merge((; kwargs...), (Pl = solver.Pl(L_tmp),))
     elseif isa(L_tmp, SparseMatrixCSC)
-        kwargs = merge((; kwargs...), (Pl = ilu(L_tmp, τ=0.01),))
+        kwargs = merge((; kwargs...), (Pl = ilu(L_tmp, τ = 0.01),))
     end
     !isnothing(solver.Pr) && (kwargs = merge((; kwargs...), (Pr = solver.Pr(L_tmp),)))
-
 
     prob = LinearProblem(L_tmp, v0)
     ρss_vec = solve(prob, solver.alg; kwargs...).u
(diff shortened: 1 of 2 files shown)


albertomercurio:
Great! I think that the most important thing is to make the code array-agnostic, that's what you have already done.

The fact that some of the operations are not yet supported on GPU, doesn't mean that they won't be fixed in the future. For example, we just need to wait for **KrylovPreconditioners.jl** supports the preconditioner for `CuSparseMatrixCSC`.

A related question is: what happens if you consider a `CuSparseMatrixCSR` with preconditioner?

BTW, don't forget to add some runtests, in order to check that everything works well.

albertomercurio:
Moreover, do you see any performance downgrade using the direct solver of **LinearSolve.jl** with respect to Julia's default?

LorenzoFioroni:
Thanks for the feedback!

I've added a test for each method, comparing the steady-state density matrix to that obtained by integrating the Lindblad master equation. The comparison is done via the recently added `tracedist` function. 

Indeed, using the direct method through LinearSolve.jl results in a little downgrade of the solver's performance. I've switched back to using the original direct solver as the default choice.

Finally, I formatted the code, which I had forgotten to do before - sorry about that!

albertomercurio:
Great. You have to format also the test file. Then everything seems to be fine.

ytdHuang (approved):

Issue #95, opened 2024-05-02
Support `permute` to re-order the tensor (Kronecker) product


ytdHuang:
Support the function [`permute` in `QuTiP`](https://github.com/qutip/qutip/blob/320996bfda3a2eb361eb36cac4208990266b97fe/qutip/core/qobj.py#L1158), something like:

```julia
function permute(A::QuantumObject, order::Vector{Int}) ... end

a = Qobj(rand(2, 2))
b = Qobj(rand(3, 3))
c = Qobj(rand(4, 4))
d = Qobj(rand(5, 5))
abcd = tensor(a, b, c, d)

permute(abcd, [4, 1, 3, 2]) == tensor(d, a, c, b) # true
```

Should support for the following types of quantum object:
- [`KetQuantumObject`](https://github.com/albertomercurio/QuantumToolbox.jl/blob/main/src/quantum_object.jl#L31)
- [`BraQuantumObject`](https://github.com/albertomercurio/QuantumToolbox.jl/blob/main/src/quantum_object.jl#L16)
- [`OperatorQuantumObject`](https://github.com/albertomercurio/QuantumToolbox.jl/blob/main/src/quantum_object.jl#L46)
- [`SuperOperatorQuantumObject`](https://github.com/albertomercurio/QuantumToolbox.jl/blob/main/src/quantum_object.jl#L61)
- [`OperatorKetQuantumObject`](https://github.com/albertomercurio/QuantumToolbox.jl/blob/main/src/quantum_object.jl#L91)
- [`OperatorBraQuantumObject`](https://github.com/albertomercurio/QuantumToolbox.jl/blob/main/src/quantum_object.jl#L76)

aarontrowbridge:
the `tensor` or [`kron`](https://github.com/albertomercurio/QuantumToolbox.jl/blob/5e24d9006f96bba4b1d1169fa4cca5d20ba9d45a/src/quantum_object.jl#L637) function, seems to be only defined for Kets, Bras, and Operators, is it still necessary to write methods for the super operators

albertomercurio:
I can’t figure out what a tensor product between two superoperators should make, physically. So, I think that the definition of tensor product on the superoperators is meaningless

aarontrowbridge:
[...]

PR #151, opened 2024-05-30
adding permute function for reordering tensor products of quantum objects

aarontrowbridge:
This hopefully resolves issue #95, it doesnt handle super operators, but perhaps that can be added in the future on top of this.


ytdHuang:
Sorry that we have re-organized the source code structure yesterday. So there's a conflict compare to the main branch. 
Could you put these new functions in [`src/qobj/arithmetic_and_attributes.jl`](https://github.com/qutip/QuantumToolbox.jl/blob/main/src/qobj/arithmetic_and_attributes.jl).

src/quantum_object.jl
```diff
@@ -717,6 +717,110 @@ Quantum Object:   type=Operator   dims=[20, 20]   size=(400, 400)   ishermitian=
...
+ψ2 = fock(2, 1)
+ψ_12 = tensor(ψ1, ψ2)
+
+permute(ψ_12, [2, 1]) ≈ tensor(ψ2, ψ1)
+
+# Returns true
+```
+
+"""
+function permute end
```
ytdHuang, line 740: Maybe just put the definition of functions here, we can merge those three methods in just one line, namely
```julia
function permute(
    A::QuantumObject{<:AbstractArray{T},ObjType},
    order::AbstractVector{Int}
) where {T,ObjType<:Union{KetQuantumObject,BraQuantumObject,OperatorQuantumObject}}
    if length(order) != length(A.dims)
        throw(ArgumentError("The order vector must have the same length as the number of subsystems"))
    end
    perm = kron_permutation(order, op.dims)
    return QuantumObject(A.data[perm, perm], A.type, A.dims[order])
end
```

src/quantum_object.jl
```diff
@@ -717,6 +717,110 @@ Quantum Object:   type=Operator   dims=[20, 20]   size=(400, 400)   ishermitian=
...
+
+#  -------------------------------------------------------------
+#  helper functions
+#  -------------------------------------------------------------
+
+function kron_rep(
```
ytdHuang, line 746: I suggest this to be an internal function, can you rename this as `_kron_rep`

src/quantum_object.jl
```diff
@@ -717,6 +717,110 @@ Quantum Object:   type=Operator   dims=[20, 20]   size=(400, 400)   ishermitian=
...
+        if i != length(dims)
+            x_i .*= "_"
+        end
+        push!(xs, x_i)
+    end
+
+    return kron(xs...)
+end
+
+function kron_permutation(perm::Vector{Int}, dims::Vector{Int})
```
ytdHuang, line 765: I suggest this to be an internal function, can you rename this as `_kron_permutation`

test/quantum_objects.jl
```diff
@@ -291,4 +291,27 @@
...
+    @test tensor(bra_a, bra_b, bra_c) ≈ tensor(bra_a, tensor(bra_b, bra_c))
+    @test tensor(op_a, op_b, op_c) ≈ tensor(op_a, tensor(op_b, op_c))
+
+    ket_abc = tensor(ket_a, ket_b, ket_c)
+    bra_abc = tensor(bra_a, bra_b, bra_c)
+    op_abc = tensor(op_a, op_b, op_c)
+
+    @test permute(ket_abc, [2, 3, 1]) ≈ tensor(ket_b, ket_c, ket_a)
+    @test permute(bra_abc, [2, 3, 1]) ≈ tensor(bra_b, bra_c, bra_a)
+    @test permute(op_abc, [2, 3, 1]) ≈ tensor(op_b, op_c, op_a)
```
ytdHuang, line 316: Can you also add the tests for raising potential errors ?
For example:
```julia
permute(ket_abc, [1, 2])
permute(ket_abc, [4, 5, 6])
```

test/quantum_objects.jl
```diff
@@ -291,4 +291,27 @@
     @test typeof(SparseMatrixCSC(Md).data) == SparseMatrixCSC{Int64,Int64}
     @test typeof(SparseMatrixCSC(Ms).data) == SparseMatrixCSC{Int64,Int64}
     @test typeof(SparseMatrixCSC{ComplexF64}(Ms).data) == SparseMatrixCSC{ComplexF64,Int64}
+
+    # tensor tests
```
ytdHuang, line 295: maybe make the comment
```
# tensor and permute tests
```

src/quantum_object.jl
```diff
@@ -717,6 +717,110 @@ Quantum Object:   type=Operator   dims=[20, 20]   size=(400, 400)   ishermitian=
...
+"""
+function permute end
+
+#  -------------------------------------------------------------
+#  helper functions
+#  -------------------------------------------------------------
+
+function kron_rep(
+    dims;
+    labels="abcdefghijklmnopqrstuvwxyz"[1:length(dims)]
```
ytdHuang, line 748: Is there a better way to define these labels for arbitrary `length(dims)`?
Because it seems that this doesn't work if the number of sub-systems is larger than `26` (`a` ~ `z`)

ytdHuang:
Could you also put the docstring of `permute` in to the documentation API list (in the section [Qobj arithmetic and attributes](https://github.com/qutip/QuantumToolbox.jl/blob/main/docs/src/api.md?plain=1#L41))

src/quantum_object.jl
```diff
@@ -717,6 +717,110 @@ Quantum Object:   type=Operator   dims=[20, 20]   size=(400, 400)   ishermitian=
...
+)
+
+    # create string representations of the subsystems
+    # and store them in the order of the permutation
+    xs = []
```
albertomercurio, line 753: We should about to create empty lists of undefined types and then pushing to them. This is inefficient and allocates memory

src/quantum_object.jl
```diff
@@ -717,6 +717,110 @@ Quantum Object:   type=Operator   dims=[20, 20]   size=(400, 400)   ishermitian=
...
+end
+
+function permute(
+    op::QuantumObject{<:AbstractArray{T}, OperatorQuantumObject},
+    order::AbstractVector{Int}
+) where T
+    if length(order) != length(op.dims)
+        throw(ArgumentError("The order vector must have the same length as the number of subsystems"))
+    end
+    perm = kron_permutation(order, op.dims)
```
albertomercurio, line 820: I don’t know how much inefficient is calling `data[perm, perm]`, especially for sparse matrices

PR #155, opened 2024-06-03
Attribute functions operating on `Qobj`

### LINKED ISSUE #148: Attribute functions operating on `Qobj`
These are the missing `Qobj.attribute` functions in `qutip`.

TODO:
- addressed by #150
  - [x] Move all these functions into a new file `qobj/arithmetic_and_attributes.jl`
  - [x] For those functions that are not attributes in `qutip`, move to a new file `qobj.functions.jl`
  - [x] make `tr` return real value if the input `Qobj` is Hermitian

- addressed by #155
  - [x] Make sure all existing functions have docstring
  - [x] generalized `dot`, same as `matrix_element`
  - [x] `isherm`
  - [x] `diag`
  - [x] `expm`
  - [x] `logm`
  - [x] `purity`
  - [x] `proj`
  - [x] `trans`
  - [x] `unit`

ytdHuang:
close #148 

This PR supports most of the attribute functions in Python `qutip.Qobj` class.

Summary of this PR:
- add missing `@doc raw` in front of each docstirings.
- add missing docstrings for several `LinearAlgebra` functions
- always throw `DimensionMismatch` exception if `A.dims != B.dims`
- introduce `isherm` (same as `LinearAlgebra.ishermitian`)
- support generalized `LinearAlgebra.dot(i, A, j)`: returns $\langle i | A | j \rangle$
  - if `A` is in the type of `Operator`, `i` and `j` need to be `Ket`
  - if `A` is in the type of `SuperOperator`, `i` and `j` need to be `OperatorKet`
- introduce `matrix_element(i, A, j)` (same as `LinearAlgebra.dot(i, A, j)`)
- introduce `trans` (same as `LinearAlgebra.transpose`)
- introduce `dag` (same as `LinearAlgebra.adjoint`)
- introduce `unit` (same as `LinearAlgebra.normalize`)
- allow `normalize(A, p)` and `normalize!(A, p)` to normalize the `QuantumObject` so that `p`-norm of `A` equals to unity. Support for the following types of `QuantumObject`:
  - If `A` is `Ket` or `Bra`, default `p = 2`
  - If `A` is `Operator`, default `p = 1`
- introduce `expm` (same as `LinearAlgebra.exp`)
- support `LinearAlgebra.sin` and `LinearAlgebra.cos`
- only support `exp`, `expm`, `sin`, `sinm`, `cos`, `cosm` for `Operator` and `SuperOperator` types of `QuantumObject` as input
- introduce `LinearAlgebra.diag`
- introduce `LinearAlgebra.log` and `logm` (same as `LinearAlgebra.log`)
- introduce `purity`: $\textrm{Tr}(\rho^2)$
- introduce `proj`: the projector of a `Ket` or `Bra`


docs/src/api.md
```diff
@@ -36,18 +36,41 @@ isoper
...
+LinearAlgebra.ishermitian
+isherm
+LinearAlgebra.issymmetric
+LinearAlgebra.isposdef
 ```
 
 ## [Qobj arithmetic and attributes](@id doc-API:Qobj-arithmetic-and-attributes)
 
 ```@docs
+Base.conj
```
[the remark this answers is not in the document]
ytdHuang: But I thought this is an attribute function @@

In `Python`, they make conjugate of `Qobj` by calling `Qobj.conj()`

docs/src/api.md
```diff
@@ -36,25 +36,53 @@ isoper
...
+LinearAlgebra.log
+LinearAlgebra.exp
 LinearAlgebra.tr
 LinearAlgebra.svdvals
 LinearAlgebra.norm
+LinearAlgebra.normalize
+LinearAlgebra.normalize!
+LinearAlgebra.inv
+LinearAlgebra.diag
+sinm
```
albertomercurio, line 62: Can you convert also `sinm` and `cosm to `LinearAlgebra.sin` and `LinearAlgebra.cos` and them put `sinm` and `cosm` as synonyms?

docs/src/api.md
```diff
@@ -36,25 +36,53 @@ isoper
...
+permute
 tidyup
 tidyup!
 get_data
 get_coherence
 partial_transpose
 ```
 
+## [Synonyms for Qobj attribute functions](@id doc-API:Synonyms-for-Qobj-attribute-functions)
+```@docs
```
albertomercurio, line 76: I think that also `isherm` should be moved here

docs/src/api.md
ytdHuang: Ohh, I didn't notice that there is also matrix `sin` and `cos` in `LinearAlgebra`.
Will implement it right away !

ytdHuang: This could be more efficient compared to the original way

docs/src/api.md
ytdHuang: sure, thanks for the reminder

docs/src/api.md
albertomercurio: No I don’t think that it will work with the LinearAlgebra functions, because we have sparse matrices, and (at least at the time) there wasn’t an implementation for such functions. That’s why I implemented the `exp` function by myself. When SparseArrays will export their `exp` `sin` `cos` functions, we can delete our definition of `exp` function.

What I mean here is to do the same as the `exp` case also for `sin` and `cos`, and then define `sinm` and `cosm` as synonyms.

ytdHuang: Okay, just to make sure I understand correctly, you mean:
```julia
LinearAlgebra.sin(A::QuantumObject) = (exp(1im * A) - exp(-1im * A)) / 2im
sinm(A::QuantumObject) = sin(A)
```

PR #158, opened 2024-06-04
Constructing general spin-j operators

ytdHuang:
This is a PR for constructing general Spin-`j` operators (same function names within Python `qutip`), where `j` is the spin quantum numbers that can only be non-negative integer or half-integer.

Summary of this PR:

- introduce `jmat`
- introduce `spin_Jx`
- introduce `spin_Jy`
- introduce `spin_Jz`
- introduce `spin_Jm`
- introduce `spin_Jp`
- introduce `spin_J_set`
- change the definition of the following Pauli operators by using `jmat`:
  - `sigmap() = jmat(0.5, :+)`
  - `sigmam() = jmat(0.5, :-)`
  - `sigmax() = 2 * jmat(0.5, :x)`
  - `sigmay() = 2 * jmat(0.5, :y)`
  - `sigmaz() = 2 * jmat(0.5, :z)`

### DIFF
--- a/src/qobj/operators.jl
+++ b/src/qobj/operators.jl
@@ -215,7 +215,7 @@ function _jm(j::Real)
     data = sqrt.(j * (j + 1) .- m .* (m .- 1))[1:end-1]
     return spdiagm(-1 => Array{ComplexF64}(data))
 end
-_jz(j::Real) = spdiagm(0 => ComplexF64[j - k for k in 0:Int(2 * j)])
+_jz(j::Real) = spdiagm(0 => Array{ComplexF64}(j .- (0:Int(2 * j))))
 
 @doc raw"""
     spin_Jx(j::Real)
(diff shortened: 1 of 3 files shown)


src/qobj/operators.jl
```diff
@@ -139,40 +140,181 @@ julia> fock(20, 4)' * a_d * fock(20, 3)
...
+
 @doc raw"""
     sigmap()
 
 Pauli ladder operator ``\hat{\sigma}_+ = \hat{\sigma}_x + i \hat{\sigma}_y``.
+
+See also [`jmat`](@ref).
 """
-sigmap() = destroy(2)
+sigmap() = jmat(0.5, Val(:+))
```
albertomercurio, line 281: Is it more efficient that the previous definition?

ytdHuang: I think for `sigmap` and `sigmam`, are not so different compared to the original methods.

However, I do think `sigmax`, `sigmay`, and `sigmaz` would be slightly more efficient.

Also, I personally think it is a bit weird to see `sigmap() = destroy(2)` lol.

albertomercurio (approved):

PR #169, opened 2024-06-16
Functions for generating quantum operators

### LINKED ISSUE #147: Functions for generating quantum operators
There are still some functions (compare to `qutip`) unsupported yet.
- #150
  - [x] Move the existing functions and the following ones into an independent file `qobj/operators.jl`
- #156
  - [x] `commutator`
  - [x] `fdestroy`
  - [x] `fcreate`
- #158
  - [x] `jmat`
  - [x] `spin_Jx`
  - [x] `spin_Jy`
  - [x] `spin_Jz`
  - [x] `spin_Jm`
  - [x] `spin_Jp`
  - [x] `spin_J_set`
- #169 
  - [x] `displace`
  - [x] `num`
  - [x] `position`
  - [x] `momentum`
  - [x] `phase`
  - [x] `squeeze`
  - [x] `tunneling`
  - [x] `qft`
- Decide not to implement at this moment
  - `identity`: this will cause conflict with `Base.identity`
  - `qdiags`: this can be implement easily using `Qobj` together with `LinearAlgebra.diagm` or `SparseArrays.spdiagm` in `Julia`
  - `squeezing`: I'm not  sure whether the current definition for generalized squeeze operator in `QuTiP` is common in the literature. So I decide not to implement it for now.
  - `charge`: I didn't find a common definition of this, maybe this can be implemented if someone needs it in the future.

ytdHuang:
close #147 

Summarize of this PR:
- introduce `displace` (and call it within `coherent`)
- introduce `num`
- introduce `position`
  - the function is not exported because it causes conflict with `Base.position`
- introduce `momentum`
  - the function is not exported because I think usually `position` and `momentum` are pairs.
- introduce `phase`
- introduce `squeeze`
- introduce `tunneling`
- introduce `qft`

### DIFF
--- a/src/qobj/operators.jl
+++ b/src/qobj/operators.jl
@@ -9,6 +9,7 @@ export displace, squeeze, num, position_op, momentum_op, phase
 export fdestroy, fcreate
 export commutator
 export tunneling
+export qft
 
 @doc raw"""
     commutator(A::QuantumObject, B::QuantumObject; anti::Bool=false)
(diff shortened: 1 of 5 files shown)


albertomercurio:
Mmmm… Yes, It seems the only option for `position`. Another option would be to not exporting it.

ytdHuang:
> Mmmm… Yes, It seems the only option for `position`. Another option would be to not exporting it. I don’t know which one is better.

I also think of the same thing in the beginning. But it would be a bit weird if `momentum` is exported and `position` is not

src/qobj/operators.jl
```diff
@@ -310,3 +413,58 @@ end
...
+        return QuantumObject(spdiagm(m => data, -m => data); type = Operator, dims = [N])
+    else
+        return QuantumObject(diagm(m => data, -m => data); type = Operator, dims = [N])
+    end
+end
+
+@doc raw"""
+    qft(dimensions)
+
+Generates a discrete Fourier transfor matrix ``\hat{F}_N`` for [Quantum Fourier Transform (QFT)](https://en.wikipedia.org/wiki/Quantum_Fourier_transform) with given argument `dimensions`.
```
albertomercurio, line 442: transform typo

ytdHuang: okay, fixed

albertomercurio (approved):

PR #172, opened 2024-06-20
introduce intrinsic constant tuple `DEFAULT_ODE_SOLVER_OPTIONS`

ytdHuang:
The default keyword arguments for ODE solvers are:
- `abstol = 1e-8`
- `reltol = 1e-6`
- `save_everystep = false` : this can avoid storing all the internal states in `sol.u` if users set `saveat = []`
- `save_end = true`

This PR also add some runtests for `sesolve` and `mesolve`.

### DIFF
--- a/src/time_evolution/time_evolution.jl
+++ b/src/time_evolution/time_evolution.jl
@@ -3,6 +3,8 @@ export TimeEvolutionSol, TimeEvolutionMCSol
 
 export liouvillian, liouvillian_floquet, liouvillian_generalized
 
+const _Default_ODE_Solver_Options = (abstol = 1e-8, reltol = 1e-6, save_everystep = false, save_end = true)
+
 @doc raw"""
     struct TimeEvolutionSol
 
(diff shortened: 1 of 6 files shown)


src/time_evolution/time_evolution.jl
```diff
@@ -3,6 +3,8 @@ export TimeEvolutionSol, TimeEvolutionMCSol
 
 export liouvillian, liouvillian_floquet, liouvillian_generalized
 
+const _Default_ODE_Solver_Options = (abstol = 1e-8, reltol = 1e-6, save_everystep = false, save_end = true)
```
albertomercurio, line 6: I would call it `DEFAULT_ODE_SOLVER_OPTIONS` so that it is clear it is something different from anything else in the package.

ytdHuang: okay, I will modify it later.

ytdHuang: Done~

albertomercurio (approved):

PR #174, opened 2024-06-21
Fix get_coherence for density matrices

albertomercurio:
The previous definition of `get_coherence` was valid only for pure states, although it was defined also for operators. Now it works also for density matrices.

### DIFF
--- a/src/qobj/arithmetic_and_attributes.jl
+++ b/src/qobj/arithmetic_and_attributes.jl
@@ -624,7 +624,7 @@ function get_coherence(ψ::QuantumObject{<:AbstractArray,KetQuantumObject})
 end
 
 function get_coherence(ρ::QuantumObject{<:AbstractArray,OperatorQuantumObject})
-    a = destroy(prod(ψ.dims))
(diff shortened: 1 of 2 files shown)


src/qobj/arithmetic_and_attributes.jl
```diff
@@ -611,23 +611,26 @@ get_data(A::QuantumObject) = A.data
...
+function get_coherence(ψ::QuantumObject{<:AbstractArray,KetQuantumObject})
+    a = destroy(prod(ψ.dims))
     α = expect(a, ψ)
     D = exp(α * a' - conj(α) * a)
 
     return α, D' * ψ
 end
 
+function get_coherence(ρ::QuantumObject{<:AbstractArray,OperatorQuantumObject})
+    a = destroy(prod(ψ.dims))
```
ytdHuang, line 627: I think its `destroy(prod(ρ.dims))` here

albertomercurio: Yes, there was a typo. Now it should be ok.

ytdHuang (approved):

PR #175, opened 2024-06-23
Fix progressbar end print

albertomercurio:
With the last improvement of the `progress_bar`, the output was very often incomplete in the final print, because it was not triggered depending on the speed of the execution.

Here I have introduced another condition that checks if the **total** time exceeds the `interval` variable. If yes, then proceed with printing.

### DIFF
--- a/src/progress_bar.jl
+++ b/src/progress_bar.jl
@@ -29,7 +29,7 @@ function next!(p::ProgressBar, io::IO = stdout)
     previous_time = p.previous_time
     interval = p.interval
 
-    (time() - previous_time) < interval && return
+    ((time() - previous_time) < interval || (time() - start_time) < interval) && return
 
     p.previous_time = time()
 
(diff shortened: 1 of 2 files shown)


albertomercurio:
Wait, I have to do one small change.

src/progress_bar.jl
```diff
@@ -29,7 +29,7 @@ function next!(p::ProgressBar, io::IO = stdout)
     previous_time = p.previous_time
     interval = p.interval
 
-    (time() - previous_time) < interval && return
+    ((time() - previous_time) < interval || (time() - start_time) < interval) && return
```
ytdHuang, line 32: should we add `return nothing` here?
To make the code more understandable

ytdHuang: and also in line 19, line 23

albertomercurio: What does `flush` return? We should return the same type. If it returns nothing, yes, we can also return nothing. However, using just `return` is often used in C, so it is not something new.

ytdHuang (approved):

PR #178, opened 2024-06-26
Resolve type instabilities when generating QuantumObjects

### LINKED ISSUE #177: Type instabilities when generating a `QuantumObject`
The simple code below fails when running the type-inference test

```julia
using Test
using QuantumToolbox

A = sprand(ComplexF64, 100, 100, 0.1)

@inferred QuantumObject(A)
```
```julia
return type QuantumObject{SparseMatrixCSC{ComplexF64, Int64}, OperatorQuantumObject} does not match inferred return type Union{QuantumObject{SparseMatrixCSC{ComplexF64, Int64}, BraQuantumObject}, QuantumObject{SparseMatrixCSC{ComplexF64, Int64}, KetQuantumObject}, QuantumObject{SparseMatrixCSC{ComplexF64, Int64}, OperatorQuantumObject}}
```

I think that we should fix it using multiple dispatch when calling it.

albertomercurio:
This PR fixes #177.

here I used the multiple dispatch to manage the different types, instead of using several `if` conditions inside a single function. The only “dowside” (if we want to call it like that), is that the generation of `BraQuantumObject` and `OperatorBraQuantumObject` needs a vector instead of a transposed vector (actually a matrix). It is then conjugate-transposed inside the function.

I added a test of the type `@inferred code…`, just to test that there are no type instabilities.

### DIFF
--- a/src/qobj/quantum_object.jl
+++ b/src/qobj/quantum_object.jl
@@ -141,6 +141,82 @@ struct QuantumObject{MT<:AbstractArray,ObjType<:QuantumObjectType} <: AbstractQu
     dims::Vector{Int}
 end
 
+function QuantumObject(
+    A::AbstractMatrix{T};
+    type::ObjType = nothing,
+    dims = nothing,
+) where {T,ObjType<:Union{Nothing,QuantumObjectType}}
+
+    if type isa Nothing
+        type = Operator # default type
+    end
+
+    if type != Operator && type != SuperOperator
+        throw(ArgumentError("The argument type must be OperatorQuantumObject or SuperOperatorQuantumObject if the input array is a matrix."))
+    end
+
+    _size = size(A)
+
+    if dims isa Nothing
+        if type isa OperatorQuantumObject
+            dims = [_size[1]]
+        elseif type isa SuperOperatorQuantumObject
+            dims = [isqrt(_size[1])]
+        end
+    else
+        _check_dims(dims)
+    end
+
+    _check_QuantumObject(type, dims, _size[1], _size[2])
+    return QuantumObject(A, type, dims)
+end
+
+function QuantumObject(
+    A::AbstractVector{T};
+    type::ObjType = nothing,
+    dims = nothing,
+) where {T,ObjType<:Union{Nothing,QuantumObjectType}}
+
+    if type isa Nothing
+        type = Ket # default type
+    end
+
+    if type != Ket && type != Bra && type != OperatorKet && type != OperatorBra
+        throw(ArgumentError("The argument type must be KetQuantumObject, BraQuantumObject, OperatorKetQuantumObject or OperatorBraQuantumObject if the input array is a vector."))
+    end
+
+    _size = (length(A), 1)
+
+    if dims isa Nothing
+        if (type isa KetQuantumObject) || (type isa BraQuantumObject)
+            dims = [_size[1]]
+        elseif (type isa OperatorKetQuantumObject) || (type isa OperatorBraQuantumObject) 
+            dims = [isqrt(_size[1])]
+        end
+    else
+        _check_dims(dims)
+    end
+
+    if (type isa BraQuantumObject) || (type isa OperatorBraQuantumObject)
+        data = A'
+        _size = (1, length(A))
+    else
+        data = A
+    end
+
+    _check_QuantumObject(type, dims, _size[1], _size[2])
+    return QuantumObject(data, type, dims)
+end
+
+function QuantumObject(
+    A::AbstractArray{T,N};
+    type::ObjType = nothing,
+    dims = nothing,
+) where {T,N,ObjType<:Union{Nothing,QuantumObjectType}}
+
+    throw(ArgumentError("The input array must be a vector or a matrix."))
+end
+
 function Base.show(
     io::IO,
     QO::QuantumObject{<:AbstractArray{T},OpType},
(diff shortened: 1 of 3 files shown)


src/qobj/quantum_object.jl
```diff
@@ -141,6 +141,87 @@ struct QuantumObject{MT<:AbstractArray,ObjType<:QuantumObjectType} <: AbstractQu
...
+    dims = nothing,
+) where {T,ObjType<:Union{Nothing,QuantumObjectType}}
+    if type isa Nothing
+        type = Operator # default type
+    end
+
+    if type != Operator && type != SuperOperator
+        throw(
+            ArgumentError(
+                "The argument type must be OperatorQuantumObject or SuperOperatorQuantumObject if the input array is a matrix.",
```
ytdHuang, line 156: Maybe
```
The argument type must be Operator or SuperOperator if the input array is a matrix.
```

src/qobj/quantum_object.jl
```diff
@@ -141,6 +141,87 @@ struct QuantumObject{MT<:AbstractArray,ObjType<:QuantumObjectType} <: AbstractQu
...
+    dims = nothing,
+) where {T,ObjType<:Union{Nothing,QuantumObjectType}}
+    if type isa Nothing
+        type = Ket # default type
+    end
+
+    if type != Ket && type != Bra && type != OperatorKet && type != OperatorBra
+        throw(
+            ArgumentError(
+                "The argument type must be KetQuantumObject, BraQuantumObject, OperatorKetQuantumObject or OperatorBraQuantumObject if the input array is a vector.",
```
ytdHuang, line 189: Maybe
```
The argument type must be either Ket, Bra, OperatorKet or OperatorBra if the input array is a vector.
```

src/qobj/quantum_object.jl
```diff
@@ -141,6 +141,87 @@ struct QuantumObject{MT<:AbstractArray,ObjType<:QuantumObjectType} <: AbstractQu
...
+function QuantumObject(
+    A::AbstractMatrix{T};
+    type::ObjType = nothing,
+    dims = nothing,
+) where {T,ObjType<:Union{Nothing,QuantumObjectType}}
+    if type isa Nothing
+        type = Operator # default type
+    end
+
+    if type != Operator && type != SuperOperator
```
ytdHuang, line 153: Maybe
```julia
elseif !isa(type, OperatorQuantumObject) && !isa(type, SuperOperatorQuantumObject)
```

src/qobj/quantum_object.jl
```diff
@@ -141,6 +141,87 @@ struct QuantumObject{MT<:AbstractArray,ObjType<:QuantumObjectType} <: AbstractQu
...
+function QuantumObject(
+    A::AbstractVector{T};
+    type::ObjType = nothing,
+    dims = nothing,
+) where {T,ObjType<:Union{Nothing,QuantumObjectType}}
+    if type isa Nothing
+        type = Ket # default type
+    end
+
+    if type != Ket && type != Bra && type != OperatorKet && type != OperatorBra
```
ytdHuang, line 186: Maybe
```julia
elseif !isa(type, KetQuantumObject) && !isa(type, BraQuantumObject) && !isa(type, OperatorKetQuantumObject) && !isa(type, OperatorBraQuantumObject)
```

src/qobj/quantum_object.jl
albertomercurio: Done

src/qobj/quantum_object.jl
albertomercurio: Done

src/qobj/quantum_object.jl
albertomercurio: Done

src/qobj/quantum_object.jl
albertomercurio: Done

test/quantum_objects.jl
```diff
@@ -16,19 +22,17 @@
...
     @test_throws DomainError Qobj(a', type = SuperOperator)
-    @test_throws DomainError Qobj(a', type = OperatorKet)
+    @test_throws ArgumentError Qobj(a', type = OperatorKet)
     @test_throws DimensionMismatch Qobj(a, dims = [2])
-    @test_throws DimensionMismatch Qobj(a', dims = [2])
-    a2 = Qobj(a')
+    @test_throws DomainError Qobj(a', dims = [2])
+    a2 = Qobj(a', type = Bra)
     a3 = Qobj(a)
-    @test dag(a3) == a2
```
ytdHuang, line 31: Is this the only place we test synonym `dag` ?

ytdHuang: Yes, I think so.
Maybe we need to add another test for this synonym `dag`, or just keep `dag` here.

albertomercurio: Ok I replace it with dag

test/quantum_objects.jl
```diff
@@ -383,4 +386,22 @@
     @test_throws ArgumentError permute(bra_bdca, wrong_order2)
     @test_throws ArgumentError permute(op_bdca, wrong_order1)
     @test_throws ArgumentError permute(op_bdca, wrong_order2)
+
+    @testset "Type Inference" begin
+        for T in [Float32, Float64, ComplexF32, ComplexF64]
+            N = 25
+            a = rand(T, N)
```
ytdHuang, line 393: Maybe also add a direct test (without specifying `type`, like the one mentioned in issue #177 ) before the for-loop ?

something like
```julia
for T in [Float32, Float64, ComplexF32, ComplexF64]
    N = 25
    a = rand(T, N)
    @inferred Qobj(a)
    for type in [Ket, OperatorKet]
       ...
```

test/quantum_objects.jl
```diff
@@ -16,19 +22,18 @@
...
+    @test_throws ArgumentError Qobj(a', type = Ket)
     @test_throws DomainError Qobj(a', type = Operator)
     @test_throws DomainError Qobj(a', type = SuperOperator)
-    @test_throws DomainError Qobj(a', type = OperatorKet)
+    @test_throws ArgumentError Qobj(a', type = OperatorKet)
     @test_throws DimensionMismatch Qobj(a, dims = [2])
-    @test_throws DimensionMismatch Qobj(a', dims = [2])
-    a2 = Qobj(a')
+    @test_throws DomainError Qobj(a', dims = [2])
+    a2 = Qobj(a', type = Bra)
```
ytdHuang, line 34: I think we don't need to specify `type=Bra` here, maybe just

```julia
a2 = Qobj(a')
```

In order to test whether the function recognize a matrix with `size = (1, N)` as a `Bra`

PR #183, opened 2024-07-04
Handle `tidyup` separately for real and imaginary values

ytdHuang:
This PR calls `tidyup!` inside `tidyup`, namely
```julia
tidyup(A, tol) = tidyup!(copy(A), tol)
```

Also, use multiple dispatch for different element types `T<:Real` or `T<:Complex` for function `tidyup!`:

If `T<:Real`, then use the origin method: 
```julia
@. A = T(abs(A) > tol) * A
```

If `T<:Complex`, then use the following method:
```julia
@. A = T(abs(real(A)) > tol) * real(A) + 1im * T(abs(imag(A)) > tol) * imag(A)
```

### DIFF
--- a/src/qobj/arithmetic_and_attributes.jl
+++ b/src/qobj/arithmetic_and_attributes.jl
@@ -590,12 +590,7 @@ Given a [`QuantumObject`](@ref) `A`, check the real and imaginary parts of each 
 """
 tidyup(A::QuantumObject{<:AbstractArray{T}}, tol::T2 = 1e-14) where {T,T2<:Real} =
     QuantumObject(tidyup(A.data, tol), A.type, A.dims)
-tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol(A, tol)
-function tidyup(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
-    A2 = copy(A)
-    _DropTol!(nonzeros(A2), tol)
-    return dropzeros!(A2)
-end
+tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = tidyup!(copy(A), tol)
 
 @doc raw"""
     tidyup!(A::QuantumObject, tol::Real=1e-14)
@@ -605,15 +600,12 @@ Given a [`QuantumObject`](@ref) `A`, check the real and imaginary parts of each 
 Note that this function is an in-place version of [`tidyup`](@ref).
 """
 tidyup!(A::QuantumObject{<:AbstractArray{T}}, tol::T2 = 1e-14) where {T,T2<:Real} = (tidyup!(A.data, tol); A)
-tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol!(A, tol)
-function tidyup!(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
-    _DropTol!(nonzeros(A), tol)
+function tidyup!(A::AbstractSparseArray{T}, tol::T2 = 1e-14) where {T,T2<:Real}
+    tidyup!(nonzeros(A), tol) # tidyup A.nzval in-place (also support for CUDA sparse arrays)
     return dropzeros!(A)
 end
-
-_DropTol(A::AbstractArray{T}, tol::Real) where {T} = _DropTol!(copy(A), tol)
-_DropTol!(A::AbstractArray{T}, tol::Real) where {T<:Real} = @. A = T(abs(A) > tol) * A
-_DropTol!(A::AbstractArray{T}, tol::Real) where {T} =
+tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T<:Real,T2<:Real} = @. A = T(abs(A) > tol) * A
+tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} =
     @. A = T(abs(real(A)) > tol) * real(A) + 1im * T(abs(imag(A)) > tol) * imag(A)
 
 @doc raw"""
--- a/src/qobj/arithmetic_and_attributes.jl
+++ b/src/qobj/arithmetic_and_attributes.jl
@@ -586,23 +586,43 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} = 
 @doc raw"""
     tidyup(A::QuantumObject, tol::Real=1e-14)
 
-Removes those elements of a QuantumObject `A` whose absolute value is less than `tol`.
+Given a [`QuantumObject`](@ref) `A`, check the real and imaginary parts of each element separately. Remove the real or imaginary value if its absolute value is less than `tol`.
 """
 tidyup(A::QuantumObject{<:AbstractArray{T}}, tol::T2 = 1e-14) where {T,T2<:Real} =
     QuantumObject(tidyup(A.data, tol), A.type, A.dims)
-tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = @. T(abs(A) > tol) * A
-tidyup(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real} = droptol!(copy(A), tol)
+tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol(A, tol)
+function tidyup(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
+    A2 = copy(A)
+    _DropTol!(A2.nzval, tol)
+    return dropzeros!(A2)
+end
 
 @doc raw"""
     tidyup!(A::QuantumObject, tol::Real=1e-14)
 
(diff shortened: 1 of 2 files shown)


src/qobj/arithmetic_and_attributes.jl
```diff
@@ -586,23 +586,43 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} =
...
+Given a [`QuantumObject`](@ref) `A`, check the real and imaginary parts of each element separately. Remove the real or imaginary value if its absolute value is less than `tol`.
 """
 tidyup(A::QuantumObject{<:AbstractArray{T}}, tol::T2 = 1e-14) where {T,T2<:Real} =
     QuantumObject(tidyup(A.data, tol), A.type, A.dims)
-tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = @. T(abs(A) > tol) * A
-tidyup(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real} = droptol!(copy(A), tol)
+tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol(A, tol)
+function tidyup(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
+    A2 = copy(A)
+    _DropTol!(A2.nzval, tol)
```
albertomercurio, line 596: calling `A2.nzval` won’t work for CUDA arrays, where they used a different name. I remember that there is a specific function (`nonzeros(A2)`?) for that, which works using multiple dispatch.

src/qobj/arithmetic_and_attributes.jl
```diff
@@ -586,23 +586,43 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} =
...
-tidyup!(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real} = droptol!(A, tol)
+tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol!(A, tol)
+function tidyup!(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
+    _DropTol!(A.nzval, tol)
+    return dropzeros!(A)
+end
+
+_DropTol(A::AbstractArray{T}, tol::Real) where {T<:Real} = @. T(abs(A) > tol) * A
+function _DropTol(A::AbstractArray{T}, tol::Real) where {T}
+    ReA = real(A)
```
albertomercurio, line 616: I’m not sure we really need to preallocate the real and imaginary parts. Since we are using the broadcasting, we can avoid the allocation and using everything inside the broadcasting. Something like

```julia
return @. T(abs(real(A)) > tol) * real(A) + 1im * T(abs(imag(A)) > tol) * imag(A)
```

It coul be slightly slower (I expect not that much), but with less memory allocations.

src/qobj/arithmetic_and_attributes.jl
```diff
@@ -586,23 +586,43 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} =
...
+    ImA = imag(A)
+    return @. T(abs(ReA) > tol) * ReA + 1im * T(abs(ImA) > tol) * ImA
+end
+_DropTol!(A::AbstractArray{T}, tol::Real) where {T<:Real} = @. A = T(abs(A) > tol) * A
+function _DropTol!(A::AbstractArray{T}, tol::Real) where {T}
+    ReA = real(A)
+    ImA = imag(A)
+    return @. A = T(abs(ReA) > tol) * ReA + 1im * T(abs(ImA) > tol) * ImA
```
albertomercurio, line 624: Same here.

src/qobj/arithmetic_and_attributes.jl
```diff
@@ -586,23 +586,43 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} =
...
 tidyup!(A::QuantumObject{<:AbstractArray{T}}, tol::T2 = 1e-14) where {T,T2<:Real} = (tidyup!(A.data, tol); A)
-tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = @. A = T(abs(A) > tol) * A
-tidyup!(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real} = droptol!(A, tol)
+tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol!(A, tol)
+function tidyup!(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
+    _DropTol!(A.nzval, tol)
+    return dropzeros!(A)
+end
+
+_DropTol(A::AbstractArray{T}, tol::Real) where {T<:Real} = @. T(abs(A) > tol) * A
```
albertomercurio, line 614: Should we use `DropTol(A, tol) = DropTol!(copy(A), tol)`?

src/qobj/arithmetic_and_attributes.jl
ytdHuang: Done~

src/qobj/arithmetic_and_attributes.jl
ytdHuang: Done~

src/qobj/arithmetic_and_attributes.jl
ytdHuang: Yes~
I almost forgot that it is broadcasting `@.` here.

src/qobj/arithmetic_and_attributes.jl
ytdHuang: Done

src/qobj/arithmetic_and_attributes.jl
```diff
@@ -586,23 +586,35 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} =
...
     tidyup(A::QuantumObject, tol::Real=1e-14)
 
-Removes those elements of a QuantumObject `A` whose absolute value is less than `tol`.
+Given a [`QuantumObject`](@ref) `A`, check the real and imaginary parts of each element separately. Remove the real or imaginary value if its absolute value is less than `tol`.
 """
 tidyup(A::QuantumObject{<:AbstractArray{T}}, tol::T2 = 1e-14) where {T,T2<:Real} =
     QuantumObject(tidyup(A.data, tol), A.type, A.dims)
-tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = @. T(abs(A) > tol) * A
-tidyup(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real} = droptol!(copy(A), tol)
+tidyup(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol(A, tol)
```
albertomercurio, line 593: Should we also use `tidyup(A, tol) = tidyup!(copy(A), tol)`?

In this way we converge the core code in only few lines.

src/qobj/arithmetic_and_attributes.jl
```diff
@@ -586,23 +586,35 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} =
...
+function tidyup(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
```
albertomercurio, line 594: Same here.

src/qobj/arithmetic_and_attributes.jl
```diff
@@ -586,23 +586,35 @@ purity(ρ::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}) where {T} =
...
 tidyup!(A::QuantumObject{<:AbstractArray{T}}, tol::T2 = 1e-14) where {T,T2<:Real} = (tidyup!(A.data, tol); A)
-tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = @. A = T(abs(A) > tol) * A
-tidyup!(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real} = droptol!(A, tol)
+tidyup!(A::AbstractArray{T}, tol::T2 = 1e-14) where {T,T2<:Real} = _DropTol!(A, tol)
+function tidyup!(A::AbstractSparseMatrix{T}, tol::T2 = 1e-14) where {T,T2<:Real}
+    _DropTol!(nonzeros(A), tol)
+    return dropzeros!(A)
+end
+
+_DropTol(A::AbstractArray{T}, tol::Real) where {T} = _DropTol!(copy(A), tol)
```
albertomercurio, line 614: So we don’t need the `_DropTol` function if we use my suggestion.

src/qobj/arithmetic_and_attributes.jl
ytdHuang: Done~

src/qobj/arithmetic_and_attributes.jl
ytdHuang: Done~

src/qobj/arithmetic_and_attributes.jl
ytdHuang: Done~

albertomercurio (approved):

PR #184, opened 2024-07-07
Fix missing `p` for in-place `normalize`

ytdHuang:
As title, the argument `p` is missing in `normalize!`

### DIFF
--- a/src/qobj/arithmetic_and_attributes.jl
+++ b/src/qobj/arithmetic_and_attributes.jl
@@ -325,9 +325,9 @@ Also, see [`norm`](@ref) about its definition for different types of [`QuantumOb
 LinearAlgebra.normalize!(
     A::QuantumObject{<:AbstractArray{T},ObjType},
     p::Real = 2,
-) where {T,ObjType<:Union{KetQuantumObject,BraQuantumObject}} = (rmul!(A.data, 1 / norm(A)); A)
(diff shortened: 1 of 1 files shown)


src/qobj/arithmetic_and_attributes.jl
```diff
@@ -325,9 +325,9 @@ Also, see [`norm`](@ref) about its definition for different types of [`QuantumOb
 LinearAlgebra.normalize!(
     A::QuantumObject{<:AbstractArray{T},ObjType},
     p::Real = 2,
-) where {T,ObjType<:Union{KetQuantumObject,BraQuantumObject}} = (rmul!(A.data, 1 / norm(A)); A)
+) where {T,ObjType<:Union{KetQuantumObject,BraQuantumObject}} = (rmul!(A.data, 1 / norm(A, p)); A)
 LinearAlgebra.normalize!(A::QuantumObject{<:AbstractArray{T},OperatorQuantumObject}, p::Real = 1) where {T} =
-    (rmul!(A.data, 1 / norm(A)); A)
+    (rmul!(A.data, 1 / norm(A, p)); A)
```
albertomercurio, line 330: Why don't we simply write `normalize!(A, p)` instead of `rmul!(A, 1/norm(A, p))`?

albertomercurio: Ok this is false because we are using our **custom** `norm` function.

albertomercurio (approved):

PR #185, opened 2024-07-07
Introduce `rand_ket` and generate `rand_dm` using Ginibre ensemble method

ytdHuang:
This PR introduces `rand_ket`, and also modify the method for generating random density matrix `rand_dm`.

I think the current method for generating random density matrix is not random enough. Because the real value of off-diagonal elements are always positive. 

Here, I implement the method to generate random density matrix from ***Ginibre ensemble***, which is also the default method in `qutip`. I think this method is the more standard one in the literature.

Furthermore, with this method, we can specify the `rank` of the density matrix, i.e., the number of eigenvalues which are positive. The other eigenvalues will be very close to zero (`< 1e-15`).

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -47,3 +47,5 @@ end
 
 _get_dense_similar(A::AbstractArray, args...) = similar(A, args...)
 _get_dense_similar(A::AbstractSparseMatrix, args...) = similar(nonzeros(A), args...)
+
+_Ginibre_ensemble(n::Int, rank::Int = n) = (randn(n, rank) + 1.0im * randn(n, rank)) / sqrt(2 * n)
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -48,4 +48,4 @@ end
 _get_dense_similar(A::AbstractArray, args...) = similar(A, args...)
 _get_dense_similar(A::AbstractSparseMatrix, args...) = similar(nonzeros(A), args...)
 
-_Ginibre_ensemble(n::Int, rank::Int = n) = (randn(n, rank) + 1.0im * randn(n, rank)) / sqrt(2 * n)
+_Ginibre_ensemble(n::Int, rank::Int = n) = randn(ComplexF64, n, rank) / sqrt(n)
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -101,6 +101,7 @@ zero_ket
 fock
 basis
 coherent
+rand_ket
 fock_dm
 coherent_dm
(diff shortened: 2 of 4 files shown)


src/utilities.jl
```diff
@@ -47,3 +47,5 @@ end
 
 _get_dense_similar(A::AbstractArray, args...) = similar(A, args...)
 _get_dense_similar(A::AbstractSparseMatrix, args...) = similar(nonzeros(A), args...)
+
+_Ginibre_ensemble(n::Int, rank::Int = n) = (randn(n, rank) + 1.0im * randn(n, rank)) / sqrt(2 * n)
```
albertomercurio, line 51: Why don't we simply write `randn(ComplexF64, n, rank)`?

albertomercurio (commented):
Is this method equal to the current one in the case `rank =N`?

src/utilities.jl
albertomercurio: Moreover, do we really need the `/ sqrt(2 * n)` if we then normalize for the trace?

ytdHuang:
> Is this method equal to the current one in the case `rank =N`?

Not equal, as I mentioned in above. I think the current method is not random enough.

albertomercurio:
In the case `rank = n` the Ginibre generates a square complex random matrix and then multiplies it by its adjoint. Like the previous implementation, I think. It is just to understand. The implementation is good, aside from the two comments made above.

src/utilities.jl
ytdHuang: The ***Ginibre ensemble*** is a set of random matrices with the entries chosen independently. 

Take full rank case (`rank = N`) as example, each entry of a $N\times N$ matrix is a complex number, with both the real and imaginary part are ***separately*** sampled from a normal distribution of mean zero and variance  $1/2N$.

The [circular law](https://en.wikipedia.org/wiki/Circular_law) states that when $N$ is large, with high probability, almost all the eigenvalues lie in the ***unit disk***. Moreover, they tend to be nearly uniformly distributed there.

From [this reference](http://dx.doi.org/10.1063/1.3595693), they explain the method to generate random density matrix from Ginibre Ensemble.

I tried `randn(ComplexF64, n, rank)` before, but it didn't give me correct Ginibre ensemble.

For the division of $\sqrt{2N}$, I keep this because we might use Ginibre ensemble somewhere else in the future (to generate other random stuffs instead of density matrix), so I thought it would be better to stick to the most original definition.

albertomercurio (approved):

PR #189, opened 2024-07-23
Fix type instabilities for `sesolve`

albertomercurio:
This is the first of a long series of Pull Requests for fixing all the type instabilities of the functions.

Here, I have downgraded the maximum version of DiffEqCallbacks.jl to v3.1.0, since after that version it generates type instabilities. Moreover, I made some small changes to the code.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,7 +28,7 @@ QuantumToolboxCUDAExt = "CUDA"
 [compat]
 ArrayInterface = "6, 7"
 CUDA = "5"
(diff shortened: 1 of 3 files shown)


albertomercurio:
It seems that it is using an old version of DiffEqCallbacs. Very strange.

albertomercurio:
I make the type instabilities test only vor versions equal or larger than 1.10. I don't know why it fails for smaller versions

ytdHuang (approved):

PR #193, opened 2024-07-24
Type inference tests mesolve

albertomercurio:
Same as PR #191 but with `mesolve`


albertomercurio:
I close this PR since this also includes changes from the #191. Once it will be merged, I will make again a PR

PR #195, opened 2024-07-25
Type inference tests mesolve

albertomercurio:
Same as of #191 but with `mesolve`


albertomercurio:
It’s very strange that the benchmark failed. I tried it in my computer and it gives the same timings. So I think it is a problem of the GitHub side.

albertomercurio:
Ok now everything is fine

ytdHuang (approved):

PR #203, opened 2024-08-13
introduce `rand_unitary` and `isunitary`

ytdHuang:
This PR introduces two functions which are the same ones in `qutip`:
- `rand_unitary`: Returns a random unitary with two different provided methods.
- `isunitary`: Test whether a `Qobj` $$U$$ is unitary operator. This function calls `Base.isapprox` to test whether $$U U^\dagger$$ is approximately equal to identity operator.

### DIFF
--- a/src/qobj/operators.jl
+++ b/src/qobj/operators.jl
@@ -43,7 +43,7 @@ function rand_unitary(dimensions::Vector{Int}, ::Val{:haar})
     # Because inv(Λ) ⋅ R has real and strictly positive elements, Q · Λ is therefore Haar distributed.
     Λ = diag(R) # take the diagonal elements of R
     Λ ./= abs.(Λ) # rescaling the elements
-    return QuantumObject(Q * Diagonal(Λ); type = Operator, dims = dimensions)
+    return QuantumObject(dense_to_sparse(Q * Diagonal(Λ)); type = Operator, dims = dimensions)
 end
 function rand_unitary(dimensions::Vector{Int}, ::Val{:exp})
     N = prod(dimensions)
--- a/src/qobj/operators.jl
+++ b/src/qobj/operators.jl
@@ -2,6 +2,7 @@
 Functions for generating (common) quantum operators.
 =#
 
+export rand_unitary
 export jmat, spin_Jx, spin_Jy, spin_Jz, spin_Jm, spin_Jp, spin_J_set
 export sigmam, sigmap, sigmax, sigmay, sigmaz
 export destroy, create, eye, projection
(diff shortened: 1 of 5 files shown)


src/qobj/operators.jl
```diff
@@ -11,6 +12,52 @@ export commutator
...
+The `distribution` specifies which of the method used to obtain the unitary matrix:
+- `:haar`: Haar random unitary matrix using the algorithm from reference 1
+- `:exp`: Uses ``\exp(-iH)``, where ``H`` is a randomly generated Hermitian operator.
+
+# References
+1. [F. Mezzadri, How to generate random matrices from the classical compact groups, arXiv:math-ph/0609050 (2007)](https://arxiv.org/abs/math-ph/0609050)
+"""
+rand_unitary(dimensions::Int, distribution::Symbol = :haar) = rand_unitary([dimensions], Val(distribution))
+rand_unitary(dimensions::Vector{Int}, distribution::Symbol = :haar) = rand_unitary(dimensions, Val(distribution))
+function rand_unitary(dimensions::Vector{Int}, ::Val{:haar})
```
albertomercurio, line 33: I have never seen changing a method using a Val of a symbol. It is ok, but maybe we can use some structs instead? In this way we leave the same consistency we changing the method of some process, like wigner, spectrum, steadystate and any other. What do you think?

albertomercurio, line 33: I have never seen changing a method using a Val of a symbol. It is ok, but maybe we can use some structs instead? In this way we leave the same consistency we changing the method of some process, like wigner, spectrum, steadystate and any other. What do you think?

ytdHuang: Actually, what I thought in the beginning is just using standard strings... (like `"haar"` or `"exp"` here), so that it can be the same as python qutip.
But we cannot use `Val{string}` to do multiple dispatch, so I decided to use `symbol`.

I think it will be a bit redundant to define structures and also the corresponding (object) constant for all these different methods. 

This might also cause troubles in the future. Because all the name of these methods would need to be exported. For example, exporting the method keyword `exp` here will cause conflicts with the `exp()` function. Then we need to give a new name for it, and that means the method name cannot be the same as python qutip.

albertomercurio: Okok, it is ok to me to do this approach. We need to keep the same syntax for all the package, so we have to change also the other methods for the other functions like spectrum, wigner, and so on..

Also, I was wondering about type inference for using this approach of `Val{:symbol}`. Could you add some type inference tests?

ytdHuang: I have added the inference tests.

I think it's not necessary to change the "solving methods", cause some solvers need extra arguments, it will be better to still use structure to save those arguments.

test/states_and_operators.jl
```diff
@@ -189,16 +189,38 @@
...
+        U2 = rand_unitary(20, :exp)
+        U3 = rand_unitary([5, 5])
+        U4 = rand_unitary([5, 5], :exp)
+        @test isunitary(U1)
+        @test isunitary(U2)
+        @test isunitary(U3)
+        @test isunitary(U4)
+        @test U1.dims == U2.dims == [20]
+        @test U3.dims == U4.dims == [5, 5]
+        @inferred QuantumObject{Matrix{ComplexF64},OperatorQuantumObject} rand_unitary(10, :haar)
```
albertomercurio, line 203: Can you just put the type inference tests inside an inner `@testset`, and also remove the manual return type? It should be automatically detected.

See for example the other inference tests.

ytdHuang: Done

albertomercurio (approved):

PR #210, opened 2024-08-29
Stochastic Schrödinger equation

lgravina1997:
Adding a solver for the Stochastic Schrödinger equation. 
References are:

Gardiner - Handbook of Stochastic processes
Breuer, Petruccione - The theory of open quantum systems 
QuTip: [SSE](https://qutip.readthedocs.io/en/master/guide/dynamics/dynamics-stochastic.html)


src/QuantumToolbox.jl
```diff
@@ -29,6 +30,7 @@ import Graphs: connected_components, DiGraph
 import IncompleteLU: ilu
 import LinearMaps: LinearMap
 import OrdinaryDiffEq: OrdinaryDiffEqAlgorithm
+import DiffEqNoiseProcess: RealWienerProcess, RealWienerProcess!
```
ytdHuang, line 33: maybe put this line under `import DiffEqCallbacks`
since we mentioned that we are importing packages in alphabetical order

src/time_evolution/time_evolution.jl
```diff
@@ -95,6 +96,37 @@ function Base.show(io::IO, sol::TimeEvolutionMCSol)
...
+    n_traj::Int
+    times::TT
+    states::TS
+    expect::TE
+    expect_all::TEA
+    converged::Bool
+    alg::StochasticDiffEq.StochasticDiffEqAlgorithm
+    abstol::T
+    reltol::T
+end
```
ytdHuang, line 115: I think we don't need to specify `T<:Real`, because now it restrict the type of `abstol` and `reltol` to be exactly the same.
How about doing it in this way:
```julia
abstol::Real
reltol::Real
```
like the way we did in [`TimeEvolutionMCSol`](https://github.com/qutip/QuantumToolbox.jl/blob/main/src/time_evolution/time_evolution.jl#L64)

albertomercurio: We thought that the last option would lead to type instabilities. Do you agree?

ytdHuang: If we remove `T<:Real`, I don't think it will cause type instabilities.
Like we did in [TimeEvolutionMCSol](https://github.com/qutip/QuantumToolbox.jl/blob/main/src/time_evolution/time_evolution.jl#L64)

ytdHuang: So the whole thing becomes
```julia
struct TimeEvolutionSSESol{
    TT<:Vector{<:Real},
    TS<:AbstractVector,
    TE<:Matrix{ComplexF64},
    TEA<:Array{ComplexF64,3}
}
    n_traj::Int
    times::TT
    states::TS
    expect::TE
    expect_all::TEA
    converged::Bool
    alg::StochasticDiffEq.StochasticDiffEqAlgorithm
    abstol::Real
    reltol::Real
end
```

albertomercurio: Yeah, I agree, indeed I think we should also change it to the other strucs. Then how the compiler knows about the type of abstol and reltol?

See for example this simple code
```julia
a = destroy(10)
sol = sesolve(a' * a, fock(10, 0), range(0, 1, 1000))
@code_warntype sol.abstol
```

```julia
MethodInstance for getproperty(::TimeEvolutionSol{Vector{Float64}, Vector{QuantumObject{Vector{ComplexF64}, KetQuantumObject}}, Matrix{ComplexF64}}, ::Symbol)
  from getproperty(x, f::Symbol) @ Base Base.jl:37
Arguments
  #self#::Core.Const(getproperty)
  x::TimeEvolutionSol{Vector{Float64}, Vector{QuantumObject{Vector{ComplexF64}, KetQuantumObject}}, Matrix{ComplexF64}}
  f::Symbol
Body::Any
1 ─      nothing
│   %2 = Base.getfield(x, f)::Any
└──      return %2
```

For the moment we defined two different types, `T1` and `T2`.

src/time_evolution/ssesolve.jl
```diff
@@ -0,0 +1,407 @@
...
+- The states will be saved depend on the keyword argument `saveat` in `kwargs`.
+- If `e_ops` is specified, the default value of `saveat=[tlist[end]]` (only save the final state), otherwise, `saveat=tlist` (saving the states corresponding to `tlist`). You can also specify `e_ops` and `saveat` separately.
+- The default tolerances in `kwargs` are given as `reltol=1e-6` and `abstol=1e-8`.
+- For more details about `alg` and extra `kwargs`, please refer to [`DifferentialEquations.jl`](https://diffeq.sciml.ai/stable/)
+
+# Returns
+
+- `sol::TimeEvolutionSSESol`: The solution of the time evolution. See also [`TimeEvolutionSSESol`](@ref)
+"""
+
```
albertomercurio, line 346: There is and extra line here, and so the documentation failed.

lgravina1997: I actually think it's because TimeEvolutionSSESol was not being exported and because it was missing the docs. Im adding them now

albertomercurio: Yes, this is one, but now you have to put it into the documentation, otherwise we have an error. 

Moreover, what I was saying is that the documentation failed because the string should be exactly above the respective code. Indeed, the returned error is
```julia-repl
Error: no docs found for 'ssesolve' in `@docs` block in src/api.md:175-198
```

albertomercurio:
Merged using a clean PR in #238

PR #213, opened 2024-09-04
Normalize states for `mcsolve` solution

albertomercurio:
## Description
The current implementation was returning unnormalized states when `saveat` argument was set. This is different from qutip behavior, where the normalized states are returned.

Here I fix this difference by normalizing the states when generating the solution.


ytdHuang (approved):

ytdHuang:
LGTM !

PR #216, opened 2024-09-06
Improve c_ops handling

albertomercurio:
## Description
Here I improve the way we handle `c_ops`. In this way, the `liouvillian` function or anything related to `c_ops` should be much easier to read.


ytdHuang (approved):
LGTM

PR #217, opened 2024-09-07
Change the type of `dims` for `QuantumObject` to `SVector` for type stabilities

albertomercurio:
## Description
With this PR I made a lot of changes, because I basically changed the internal structure of the `QuantumObject` constructor.

I changed the type of the `dims` field from `Vector{Int}` to `SVector{N,Int}`, where `N` is the number of subsystems (e.g., two qubits, a qubit and a cavity….).

This change would fix possible type instabilities for functions like `ptrace`, where the temporary reshaped array depends on the length of the `dims` field.

Being a static fixed-length vector, the length of the `dims` field is already known at compile time, improving the performances.

It would also improve performances on functions dealing with `dims`, for example `dfd_mesolve`, where the Hilbert space dimension is changed dynamically during a time evolution.

> [!IMPORTANT]
> I kept the support for `Vector` as input, generating a warning string, and then converting it into a `SVector`

This PR is just a starting point. Here everything works. In the next PRs I will try to optimize the code to take advantage of the power of `SVector`s.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -28,6 +28,7 @@ const PAGES = [
             "users_guide/QuantumObject/QuantumObject.md",
             "users_guide/QuantumObject/QuantumObject_functions.md",
         ],
+        "The Importance of Type-Stability" => "users_guide/type_stability.md",
         "Manipulating States and Operators" => "users_guide/states_and_operators.md",
         "Tensor Products and Partial Traces" => "users_guide/tensor.md",
         "Time Evolution and Dynamics" => [
--- a/docs/src/users_guide/QuantumObject/QuantumObject.md
+++ b/docs/src/users_guide/QuantumObject/QuantumObject.md
@@ -39,6 +39,7 @@ Qobj(rand(4, 4))
 ```
 
 ```@example Qobj
+M = rand(ComplexF64, 4, 4)
 Qobj(M, dims = [2, 2])  # dims as Vector
 Qobj(M, dims = (2, 2))  # dims as Tuple (recommended)
 Qobj(M, dims = SVector(2, 2)) # dims as StaticArrays.SVector (recommended)
--- a/docs/src/users_guide/QuantumObject/QuantumObject.md
+++ b/docs/src/users_guide/QuantumObject/QuantumObject.md
@@ -39,8 +39,11 @@ Qobj(rand(4, 4))
 ```
 
 ```@example Qobj
-Qobj(rand(4, 4), dims = [2, 2])
+Qobj(rand(4, 4), dims = (2, 2))
 ```
+
+> [!IMPORTANT]
+> Please note that here we put the `dims` as a tuple `(2, 2)`. Although it supports also `Vector` type (`dims = [2, 2]`), it is recommended to use `Tuple` or `SVector` from [`StaticArrays.jl`](https://github.com/JuliaArrays/StaticArrays.jl) to improve performance. For a brief explanation on the impact of the type of `dims`, see the Section [The Importance of Type-Stability](@ref doc:Type-Stability).
 
 ```@example Qobj
 Qobj(rand(4, 4), type = SuperOperator)
--- a/docs/src/users_guide/QuantumObject/QuantumObject.md
+++ b/docs/src/users_guide/QuantumObject/QuantumObject.md
@@ -39,7 +39,9 @@ Qobj(rand(4, 4))
 ```
 
 ```@example Qobj
-Qobj(rand(4, 4), dims = (2, 2))
+Qobj(M, dims = [2, 2])  # dims as Vector
+Qobj(M, dims = (2, 2))  # dims as Tuple (recommended)
+Qobj(M, dims = SVector(2, 2)) # dims as StaticArrays.SVector (recommended)
 ```
 
 > [!IMPORTANT]
--- a/src/QuantumToolbox.jl
+++ b/src/QuantumToolbox.jl
@@ -32,7 +32,8 @@ import OrdinaryDiffEq: OrdinaryDiffEqAlgorithm
 import Pkg
 import Random
 import SpecialFunctions: loggamma
-import StaticArraysCore: SVector, MVector
+@reexport import StaticArraysCore: SVector
+import StaticArraysCore: MVector
 
 # Setting the number of threads to 1 allows
 # to achieve better performances for more massive parallelizations
--- a/src/QuantumToolbox.jl
+++ b/src/QuantumToolbox.jl
@@ -32,7 +32,7 @@ import OrdinaryDiffEq: OrdinaryDiffEqAlgorithm
 import Pkg
 import Random
 import SpecialFunctions: loggamma
-import StaticArrays: SVector, MVector
+import StaticArraysCore: SVector, MVector
 
 # Setting the number of threads to 1 allows
 # to achieve better performances for more massive parallelizations
--- a/src/QuantumToolbox.jl
+++ b/src/QuantumToolbox.jl
@@ -32,7 +32,7 @@ import OrdinaryDiffEq: OrdinaryDiffEqAlgorithm
 import Pkg
 import Random
 import SpecialFunctions: loggamma
-import StaticArrays: SVector
+import StaticArrays: SVector, MVector
 
 # Setting the number of threads to 1 allows
 # to achieve better performances for more massive parallelizations
--- a/src/QuantumToolbox.jl
+++ b/src/QuantumToolbox.jl
@@ -32,6 +32,7 @@ import OrdinaryDiffEq: OrdinaryDiffEqAlgorithm
 import Pkg
 import Random
 import SpecialFunctions: loggamma
+import StaticArrays: SVector
 
 # Setting the number of threads to 1 allows
 # to achieve better performances for more massive parallelizations
--- a/src/qobj/quantum_object.jl
+++ b/src/qobj/quantum_object.jl
@@ -218,27 +218,19 @@ end
 _get_size(A::AbstractMatrix) = size(A)
 _get_size(A::AbstractVector) = (length(A), 1)
 
-_check_dims(dims::Union{NTuple{N,T},SVector{N,T},MVector{N,T}}) where {N,T<:Integer} =
-    (all(>(0), dims) && length(dims) > 0) || throw(
-        DomainError(
-            dims,
-            "The argument dims must be a Tuple or a StaticVector of non-zero length and contain only positive integers.",
-        ),
-    )
-function _check_dims(dims::Vector{T}) where {T<:Integer}
-    @warn "The argument dims should be a Tuple or a StaticVector for better performance."
-    return (all(>(0), dims) && length(dims) > 0) || throw(
-        DomainError(
-            dims,
-            "The argument dims must be a Tuple or a StaticVector of non-zero length and contain only positive integers.",
-        ),
-    )
+_non_static_array_warning(argname, arg::Tuple{}) =
+    throw(ArgumentError("The argument $argname must be a Tuple or a StaticVector of non-zero length."))
+_non_static_array_warning(argname, arg::Union{SVector{N,T},MVector{N,T},NTuple{N,T}}) where {N,T} = nothing
+_non_static_array_warning(argname, arg::AbstractVector{T}) where {T} =
+    @warn "The argument $argname should be a Tuple or a StaticVector for better performance. Try to use `$argname = $(Tuple(arg))` or `$argname = SVector(" *
+          join(arg, ", ") *
+          ")` instead of `$argname = $arg`." maxlog = 1
+
+function _check_dims(dims::Union{AbstractVector{T},NTuple{N,T}}) where {T<:Integer,N}
+    _non_static_array_warning("dims", dims)
+    return (all(>(0), dims) && length(dims) > 0) ||
+           throw(DomainError(dims, "The argument dims must be of non-zero length and contain only positive integers."))
 end
-_check_dims(dims::Tuple{}) = throw(
-    ArgumentError(
-        "The argument dims must be a Tuple or a StaticVector of non-zero length and contain only positive integers.",
-    ),
-)
 _check_dims(dims::Any) = throw(
     ArgumentError(
         "The argument dims must be a Tuple or a StaticVector of non-zero length and contain only positive integers.",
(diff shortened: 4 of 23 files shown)


albertomercurio:
As an example, I was doing some benchmarks comparing the current implementation of `ptrace` and the one with `SVector`.

This is the result with `SVector`
```julia-repl
julia> g = fock(2, 1)
Quantum Object:   type=Ket   dims=[2]   size=(2,)
2-element Vector{ComplexF64}:
 0.0 + 0.0im
 1.0 + 0.0im

julia> e = fock(2, 0)
Quantum Object:   type=Ket   dims=[2]   size=(2,)
2-element Vector{ComplexF64}:
 1.0 + 0.0im
 0.0 + 0.0im

julia> state = normalize(kron(g, e) + kron(e, g))
Quantum Object:   type=Ket   dims=[2, 2]   size=(4,)
4-element Vector{ComplexF64}:
                0.0 + 0.0im
 0.7071067811865475 + 0.0im
 0.7071067811865475 + 0.0im
                0.0 + 0.0im

julia> ptrace(state, 1)
Quantum Object:   type=Operator   dims=[2]   size=(2, 2)   ishermitian=true
2×2 Matrix{ComplexF64}:
 0.5+0.0im  0.0+0.0im
 0.0+0.0im  0.5+0.0im

julia> @code_warntype ptrace(state, 1)
MethodInstance for QuantumToolbox.ptrace(::QuantumObject{Vector{ComplexF64}, KetQuantumObject, 2}, ::Int64)
  from ptrace(QO::QuantumObject, sel::Int64) @ QuantumToolbox ~/.julia/dev/QuantumToolbox/src/qobj/arithmetic_and_attributes.jl:529
Arguments
  #self#::Core.Const(QuantumToolbox.ptrace)
  QO::QuantumObject{Vector{ComplexF64}, KetQuantumObject, 2}
  sel::Int64
Body::Union{QuantumObject{Matrix{ComplexF64}, BraQuantumObject, 1}, QuantumObject{Matrix{ComplexF64}, OperatorQuantumObject, 1}}
1 ─ %1 = QuantumToolbox.SVector(sel)::StaticArraysCore.SVector{1, Int64}
│   %2 = QuantumToolbox.ptrace(QO, %1)::Union{QuantumObject{Matrix{ComplexF64}, BraQuantumObject, 1}, QuantumObject{Matrix{ComplexF64}, OperatorQuantumObject, 1}}
└──      return %2


julia> using BenchmarkTools

julia> @benchmark ptrace($state, 1)
BenchmarkTools.Trial: 10000 samples with 240 evaluations.
 Range (min … max):  306.850 ns … 208.118 μs  ┊ GC (min … max):  0.00% … 99.70%
 Time  (median):     320.371 ns               ┊ GC (median):     0.00%
 Time  (mean ± σ):   385.384 ns ±   2.132 μs  ┊ GC (mean ± σ):  11.99% ±  5.47%

   ▂▅▄█▆▁                                                        
  ▂██████▇▄▃▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▂▂▃▄▃▂▂▁ ▂
  307 ns           Histogram: frequency by time          469 ns <

 Memory estimate: 736 bytes, allocs estimate: 13.
```

And this is with the latest stable version
```julia-repl
julia> @benchmark ptrace($state, 1)
BenchmarkTools.Trial: 10000 samples with 40 evaluations.
 Range (min … max):  903.350 ns …   4.131 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):       1.137 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):     1.114 μs ± 126.638 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▃▃                              ▂▄▆▇█▇▅▄▂▁                  
  ▂▃▇██▇▄▃▂▂▂▂▂▂▂▂▂▂▁▂▁▂▁▂▁▂▂▂▂▃▃▄▅▇███████████▇▆▅▅▄▄▃▃▃▃▃▂▂▂▂▂ ▄
  903 ns           Histogram: frequency by time         1.27 μs <

 Memory estimate: 1.09 KiB, allocs estimate: 20.
```

Almost 4 times slower.

docs/make.jl
```diff
@@ -28,6 +28,7 @@ const PAGES = [
             "users_guide/QuantumObject/QuantumObject.md",
             "users_guide/QuantumObject/QuantumObject_functions.md",
         ],
+        "The Importance of Type-Stability" => "users_guide/type_stability.md",
```
ytdHuang, line 31: Maybe put this page under "Basic operations on quantum objects" ?

ytdHuang: not necessary, I just thought it might be reasonable cause the type of `dims` is part of `QuantumObject`

docs/src/users_guide/QuantumObject/QuantumObject.md
```diff
@@ -39,9 +39,12 @@ Qobj(rand(4, 4))
 ```
 
 ```@example Qobj
-Qobj(rand(4, 4), dims = [2, 2])
+Qobj(rand(4, 4), dims = (2, 2))
 ```
```
ytdHuang, line 43: maybe write this example block as:
```
Qobj(rand(ComplexF64, 4, 4), dims = [2, 2])  # dims as Vector
Qobj(rand(ComplexF64, 4, 4), dims = (2, 2))  # dims as Tuple
Qobj(rand(ComplexF64, 4, 4), dims = SVector(2, 2)) # dims as StaticArrays.SVector
```
I'm thinking of reexporting `SVector`, so that the users can directly call `SVector`.
I also add `ComplexF64` here to show the people from python that we can specify element types in julia.

src/QuantumToolbox.jl
```diff
@@ -32,6 +32,7 @@ import OrdinaryDiffEq: OrdinaryDiffEqAlgorithm
 import Pkg
 import Random
 import SpecialFunctions: loggamma
+import StaticArraysCore: SVector, MVector
```
ytdHuang, line 35: Since we mention about `SVector` a lot in documentation and also some warning messages.
I think it would be better to reexport `SVector`:
```julia
@reexport import StaticArraysCore: SVector # put this together with other re-export packages
import StaticArraysCore: MVector # put this here
```

test/quantum_objects.jl
```diff
@@ -51,19 +51,22 @@
 
     # unsupported type of dims
     @testset "unsupported dims" begin
-        @test_throws ArgumentError Qobj(rand(2, 2), dims = 2)
-        @test_throws ArgumentError Qobj(rand(2, 2), dims = [2.0])
-        @test_throws ArgumentError Qobj(rand(2, 2), dims = [2.0 + 0.0im])
-        @test_throws DomainError Qobj(rand(2, 2), dims = [0])
-        @test_throws DomainError Qobj(rand(2, 2), dims = [2, -2])
+        @test_throws ArgumentError Qobj(rand(2, 2), dims = (2.0,))
```
ytdHuang, line 54: maybe just put `dims = 2.0` ?
since we already defined `_check_dims(dims::Any)` to throw an `ArgumentError`

test/quantum_objects.jl
```diff
@@ -51,19 +51,22 @@
...
+        @test_throws ArgumentError Qobj(rand(2, 2), dims = (2.0 + 0.0im,))
+        @test_throws DomainError Qobj(rand(2, 2), dims = (0,))
```
ytdHuang, line 56: maybe just put `dims = 2.0`, `dims = 2.0 + 0.0im`, and `dims = 0` here?
since we already defined `_check_dims(dims::Any)` and `_check_dims(dims::Integer)`

test/quantum_objects.jl
```diff
@@ -51,19 +51,22 @@
...
     end
 
     @testset "Ket and Bra" begin
         N = 10
         a = rand(ComplexF64, 10)
         # @test_logs (:warn, "The norm of the input data is not one.") QuantumObject(a)
-        @test_throws DimensionMismatch Qobj(a, dims = [2])
-        @test_throws DimensionMismatch Qobj(a', dims = [2])
+        @test_throws DimensionMismatch Qobj(a, dims = (2,))
+        @test_throws DimensionMismatch Qobj(a', dims = (2,))
```
ytdHuang, line 69: same here, `dims = 2` ?

test/quantum_objects.jl
```diff
@@ -103,7 +106,7 @@
         @test isoperket(a3) == false
         @test isoperbra(a3) == false
         @test isunitary(a3) == false
-        @test_throws DimensionMismatch Qobj(a, dims = [2])
+        @test_throws DimensionMismatch Qobj(a, dims = (2,))
```
ytdHuang, line 109: same here, dims = 2 ?

test/quantum_objects.jl
```diff
@@ -137,8 +140,8 @@
         @test L * ρ_ket ≈ -1im * (+(spre(H) * ρ_ket) - spost(H) * ρ_ket)
         @test (ρ_bra * L')' == L * ρ_ket
         @test sum((conj(ρ) .* ρ).data) ≈ dot(ρ_ket, ρ_ket) ≈ ρ_bra * ρ_ket
-        @test_throws DimensionMismatch Qobj(ρ_ket.data, type = OperatorKet, dims = [4])
-        @test_throws DimensionMismatch Qobj(ρ_bra.data, type = OperatorBra, dims = [4])
+        @test_throws DimensionMismatch Qobj(ρ_ket.data, type = OperatorKet, dims = (4,))
+        @test_throws DimensionMismatch Qobj(ρ_bra.data, type = OperatorBra, dims = (4,))
```
ytdHuang, line 144: same here, `dims = 4` ?

test/states_and_operators.jl
```diff
@@ -14,8 +14,8 @@
 
     @testset "fock state" begin
         # fock, basis, and fock_dm
-        @test fock_dm(4; dims = [2, 2], sparse = true) ≈ ket2dm(basis(4; dims = [2, 2]))
-        @test_throws DimensionMismatch fock(4; dims = [2])
+        @test fock_dm(4; dims = (2, 2), sparse = true) ≈ ket2dm(basis(4; dims = (2, 2)))
+        @test_throws DimensionMismatch fock(4; dims = (2,))
```
ytdHuang, line 18: same here, `dims = 2` ?

test/states_and_operators.jl
```diff
@@ -312,9 +312,9 @@
 
     @testset "identity operator" begin
         I_op1 = qeye(4)
-        I_op2 = qeye(4, dims = [2, 2])
+        I_op2 = qeye(4, dims = (2, 2))
         I_su1 = qeye(4, type = SuperOperator)
-        I_su2 = qeye(4, type = SuperOperator, dims = [2])
+        I_su2 = qeye(4, type = SuperOperator, dims = (2,))
```
ytdHuang, line 317: same here, `dims = 2` ?

test/states_and_operators.jl
```diff
@@ -326,17 +326,17 @@
         @test (I_op2 == I_su1) == false
         @test (I_op2 == I_su2) == false
         @test (I_su1 == I_su2) == true
-        @test_throws DimensionMismatch qeye(4, dims = [2])
+        @test_throws DimensionMismatch qeye(4, dims = (2,))
```
ytdHuang, line 329: same here, `dims = 2` ?

test/states_and_operators.jl
```diff
@@ -326,17 +326,17 @@
...
+        @test_throws DimensionMismatch qeye(4, type = SuperOperator, dims = (2, 2))
     end
 
     @testset "superoperators" begin
         # spre, spost, and sprepost
         Xs = sigmax()
         Xd = sparse_to_dense(Xs)
-        A_wrong1 = Qobj(rand(4, 4), dims = [4])
-        A_wrong2 = Qobj(rand(4, 4), dims = [2, 2])
+        A_wrong1 = Qobj(rand(4, 4), dims = (4,))
```
ytdHuang, line 338: same here, `dims = 4` ?

src/qobj/quantum_object.jl
```diff
@@ -125,28 +125,43 @@ Julia struct representing any quantum objects.
...
+        _check_QuantumObject(type, dims, _size[1], _size[2])
+
+        N = length(dims)
+
+        return new{MT,ObjType,N}(data, type, SVector{N,Int}(dims))
+    end
+end
+
+function QuantumObject(A::AbstractArray, type::ObjType, dims::Integer) where {ObjType<:QuantumObjectType}
+    return QuantumObject(A, type, SVector{1,Int}(dims))
```
ytdHuang, line 156: should check whether `dims::Integer >= 1`, cause the following call `QuantumObject(A, type, SVector{1,Int}(dims))` will not be the inner constructor above, so it will not go through `_check_dims`.
That is, the following call might work:
```julia
QuantumObject(rand(2, 2), Operator, 0)
```

docs/make.jl
albertomercurio: My plan was also to discuss about type instabilities in general. For example, we have cases where it is recommended to use `Val(argument` instead of `argument` in a function. So I will make some practical examples which don't necessary involve `QuantumObject`

src/qobj/quantum_object.jl
albertomercurio: Actually this doesn't work, because `check_dims` is inside the definition of `QuantumObject. So everything calling `QuantumObject` would pass to `_check_dims`.

docs/make.jl
albertomercurio: So I decide to make a dedicated page, also for emphasize how important is type-stability for those who want maximum performances.

docs/src/users_guide/QuantumObject/QuantumObject.md
albertomercurio: Actually I don't know if it is useful to export `SVector`. I don't think that it would be used, one can always use a `Tuple` normally.

I think that people who need `SVector` they need it for other reasons, like dynamics of very small systems where `SVector` is useful.

Even if re-exported, it is much simple to write `Qobj(A, dims=(2,2))` instead of `Qobj(A, dims=SVector(2,2))`. Do you agree?

docs/make.jl
ytdHuang: okay

docs/src/users_guide/QuantumObject/QuantumObject.md
ytdHuang: as we discussed in Zulip, we can resolve this

ytdHuang (approved):

PR #221, opened 2024-09-08
Fix type instabilities for almost all functions

albertomercurio:
## Description
Here I solved several type instabilities. I added a test for any of those functions.

In the next PRs I will fix the type instabilities for `dfd_mesolve` and `dsf_mesolve`.

Just as an example, the `ptrace`  function is now 10 times faster than the one of v0.12.0. I didn't benchmark all the funcitons, but now the code is more clear and efficient.

I have also removed the dependency for `LinearMaps.jl` since it was giving a weak type instability. I created an ad-hoc linear map struct for that.


albertomercurio:
In the previous runtests the benchmark succeded, so I think this was a fluctuation.

test/quantum_objects.jl
```diff
@@ -294,40 +270,104 @@
         @test typeof(SparseMatrixCSC{ComplexF64}(Ms).data) == SparseMatrixCSC{ComplexF64,Int64}
     end
 
+    @testset "Type Inference (QuantumObject)" begin
+        if VERSION >= v"1.10"
```
ytdHuang, line 274: we don't need `VERSION >= v"1.10"` anymore

src/qobj/arithmetic_and_attributes.jl
```diff
@@ -576,14 +578,14 @@ function _ptrace_oper(QO::AbstractArray, dims::Union{SVector,MVector}, sel)
 
     ρmat = reshape(QO, reverse(vcat(dims, dims))...)
     topermute = 2 * nd + 1 .- qtrace_sel
-    ρmat = PermutedDimsArray(ρmat, reverse(topermute))
+    ρmat = permutedims(ρmat, reverse(topermute)) # TODO: use PermutedDimsArray when Julia v1.11.0 is released
```
ytdHuang, line 581: why do we have to wait until `Julia v1.11` ?

albertomercurio: Because it generates type instabilities. They fixed with a PR which will be with their release.

ytdHuang: okay!

ytdHuang (approved):

PR #229, opened 2024-09-19
Fix type conversion of `tlist` in time evolution

ytdHuang:
In GPU calculations, the arrays might be specified with 32-bit element type.

The current conversion of `tlist` always converts to `Float64`, which might cause extra memory allocation or potential errors (for supporting different GPU platforms) in the future development.

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -65,3 +65,13 @@ _non_static_array_warning(argname, arg::AbstractVector{T}) where {T} =
     @warn "The argument $argname should be a Tuple or a StaticVector for better performance. Try to use `$argname = $(Tuple(arg))` or `$argname = SVector(" *
           join(arg, ", ") *
           ")` instead of `$argname = $arg`." maxlog = 1
+
+# convert tlist in time evolution
+_convert_tlist(::Int32, tlist::AbstractVector) = _convert_tlist(Val(32), tlist)
+_convert_tlist(::Float32, tlist::AbstractVector) = _convert_tlist(Val(32), tlist)
+_convert_tlist(::ComplexF32, tlist::AbstractVector) = _convert_tlist(Val(32), tlist)
+_convert_tlist(::Int64, tlist::AbstractVector) = _convert_tlist(Val(64), tlist)
+_convert_tlist(::Float64, tlist::AbstractVector) = _convert_tlist(Val(64), tlist)
+_convert_tlist(::ComplexF64, tlist::AbstractVector) = _convert_tlist(Val(64), tlist)
+_convert_tlist(::Val{32}, tlist::AbstractVector) = convert(Vector{Float32}, tlist)
+_convert_tlist(::Val{64}, tlist::AbstractVector) = convert(Vector{Float64}, tlist)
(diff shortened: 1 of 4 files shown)


src/steadystate.jl
```diff
@@ -95,10 +95,14 @@ function steadystate(
     (H.dims != ψ0.dims) && throw(DimensionMismatch("The two quantum objects are not of the same Hilbert dimension."))
 
     N = prod(H.dims)
-    u0 = mat2vec(ket2dm(ψ0).data)
+    u0 = _convert_u0(mat2vec(ket2dm(ψ0).data))
+
+    Ftype = real(eltype(u0))
+    Tspan = (convert(Ftype, 0), convert(Ftype, tspan)) # Convert it to support GPUs and avoid type instabilities for OrdinaryDiffEq.jl
```
albertomercurio, line 101: We could just write `Ftype(0)` and `Ftype(tspan)`

src/utilities.jl
```diff
@@ -65,3 +65,6 @@ _non_static_array_warning(argname, arg::AbstractVector{T}) where {T} =
     @warn "The argument $argname should be a Tuple or a StaticVector for better performance. Try to use `$argname = $(Tuple(arg))` or `$argname = SVector(" *
           join(arg, ", ") *
           ")` instead of `$argname = $arg`." maxlog = 1
+
+# make sure u0 in time evolution is dense vector and has complex element type
+_convert_u0(u0::AbstractVector{T}) where {T<:Number} = convert(Vector{complex(T)}, u0)
```
albertomercurio, line 70: Can we use the `sparse_to_dense` function instead?

ext/QuantumToolboxCUDAExt.jl
```diff
@@ -89,4 +90,7 @@ _change_eltype(::Type{T}, ::Val{32}) where {T<:AbstractFloat} = Float32
 _change_eltype(::Type{Complex{T}}, ::Val{64}) where {T<:Union{Int,AbstractFloat}} = ComplexF64
 _change_eltype(::Type{Complex{T}}, ::Val{32}) where {T<:Union{Int,AbstractFloat}} = ComplexF32
 
+# make sure u0 in time evolution is dense vector and has complex element type
+_convert_u0(u0::Union{CuArray{T},CuSparseVector{T}}) where {T<:Number} = convert(CuArray{complex(T)}, u0)
```
albertomercurio, line 94: Also here, can we use the `sparse_to_dense` function instead?

src/steadystate.jl
ytdHuang: Done

src/qobj/quantum_object.jl
```diff
@@ -368,3 +368,7 @@ SparseArrays.SparseMatrixCSC(A::QuantumObject{<:AbstractMatrix}) =
     QuantumObject(SparseMatrixCSC(A.data), A.type, A.dims)
 SparseArrays.SparseMatrixCSC{T}(A::QuantumObject{<:SparseMatrixCSC}) where {T<:Number} =
     QuantumObject(SparseMatrixCSC{T}(A.data), A.type, A.dims)
+
+# functions for getting Float or Complex element type
+_FType(::QuantumObject{<:AbstractArray{T}}) where {T<:Number} = _FType(T)
```
albertomercurio, line 373: I think it is better to put them together with the othe definitions in the utilities file.

albertomercurio (approved):

PR #231, opened 2024-09-22
Add pipeline to run core and code quality tests under `Julia nightly`

ytdHuang:
as title


albertomercurio:
Does the run test fail is only the nightly test fails?

ytdHuang:
> Does the run test fail is only the nightly test fails?

Because the pipeline config is in separate files (`CI.yml` and `CI-Julia-nightly.yml`), so they are independent.
If the nightly tests fails, it does not affect the original runtests in `CI.yml`

albertomercurio (approved):

PR #234, opened 2024-09-24
Introduce `Makefile` to make development more convenient

ytdHuang:
This PR introduces `Makefile` to support `make` commands.

In this way, we can simplify several commands during development:

For example:
```
make docs   # build documentation
make test   # run tests
make format # format files with JuliaFormatter
```


albertomercurio (approved):
Great feature!

PR #235, opened 2024-09-26
Automatically convert TimeDependentOperatorSum to `liouvillian`

albertomercurio:
## Description
In the case of `mesolve`, the current implementation requires that `TimeEvolutionOperatorSum` contains only `SuperOperator`s. With this PR I relax this request, converting it to `SuperOperator` internally.


ytdHuang (approved):
LGTM!

PR #240, opened 2024-09-28
[Docs] Update docstring and documentation for `steadystate`

ytdHuang:
## Description
Summary of this PR:
- add missing docstrings for `steadystate` and corresponding solvers.
- update documentation for `steadystate`.

### DIFF
--- a/docs/src/users_guide/steadystate.md
+++ b/docs/src/users_guide/steadystate.md
@@ -1,0 +1,133 @@
+# [Solving for Steady-State Solutions](@id doc:Solving-for-Steady-State-Solutions)
+
+## Introduction
+
+For time-independent open quantum systems with decay rates larger than the corresponding excitation rates, the system will tend toward a steady state as ``t\rightarrow\infty`` that satisfies the equation
+
+```math
+\frac{d\hat{\rho}_{\textrm{ss}}}{dt} = \mathcal{L}\hat{\rho}_{\textrm{ss}}=0.
+```
+
+Although the requirement for time-independence seems quite restrictive, one can often employ a transformation to the interaction picture that yields a time-independent Hamiltonian. For many these systems, solving for the asymptotic density matrix ``\hat{\rho}_{\textrm{ss}}`` can be achieved using direct or iterative solution methods faster than using master equation or Monte-Carlo simulations. Although the steady state equation has a simple mathematical form, the properties of the Liouvillian operator are such that the solutions to this equation are anything but straightforward to find.
+
+## Steady State solvers in `QuantumToolbox.jl`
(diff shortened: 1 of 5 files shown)


docs/src/users_guide/steadystate.md
```diff
@@ -0,0 +1,133 @@
...
+
+# Define parameters
+N = 20  # number of basis states to consider
+a = destroy(N)
+H = a' * a
+ψ0 = basis(N, 10)  # initial state
+κ = 0.1  # coupling to oscillator
+
+# collapse operators
+c_op_list = QuantumObject[]
```
albertomercurio, line 81: Predefining an empty vector and then pushing into it is fine. However, if doing so, we should define the entire type to make it type-stable. In this case, the compile would only know that `c_op_list` is a `Vector{QuantumObject}` and not something more specific like `Vector{QuantumObject{SparseMatrixCSC{ComplexF64, Int64}, Operator, 1}`, so it would lead to type instabilities. 

I know that this is just an example, but I would always show type-stable examples. In this way, a new user that simply performs copy and paste, will have already a type stable code.

In this case we don't need to push. `\gamma` is already different from zero. Such that we can make the code evens simpler.

docs/src/users_guide/steadystate.md
```diff
@@ -0,0 +1,133 @@
...
+| **Solver** | **Description** |
+|:-----------|:----------------|
+| [`SteadyStateDirectSolver()`](@ref SteadyStateDirectSolver) | Directly solve ``Ax=b`` using the standard method given in `Julia` `LinearAlgebra` |
+| [`SteadyStateLinearSolver()`](@ref SteadyStateLinearSolver) | Directly solve ``Ax=b`` using the algorithms given in [`LinearSolve.jl`](https://docs.sciml.ai/LinearSolve/stable/) |
+| [`SteadyStateEigenSolver()`](@ref SteadyStateEigenSolver) | Find the zero (or lowest) eigenvalue of ``\mathcal{L}`` using [`eigsolve`](@ref) |
+| [`SteadyStateODESolver()`](@ref SteadyStateODESolver) | Solving time evolution with algorithms given in [`DifferentialEquations.jl` (ODE Solvers)](https://docs.sciml.ai/DiffEqDocs/stable/solvers/ode_solve/) |
+
+## Using Steady State solvers
+
+The function [`steadystate`](@ref) can take either a Hamiltonian and a list of collapse operators `c_ops` as input, generating internally the corresponding Liouvillian ``\mathcal{L}`` in Lindblad form, or alternatively, a Liouvillian ``\mathcal{L}`` passed by the user.
```
albertomercurio, line 25: I would also explain that it also supports directly calling `steadystate(L)`, where `L` is the Liouvillian.

docs/src/users_guide/steadystate.md
ytdHuang: okay !

docs/src/users_guide/steadystate.md
ytdHuang: okay !

albertomercurio (approved):

PR #243, opened 2024-09-28
Replace `n_traj` with `ntraj` for QuTiP compatibility

albertomercurio:
## Description
QuTiP uses `ntraj` as an argument of Monte Carlo and stochastic simulations. Here I change the current argument from `n_traj` to `ntraj`, to align to QuTiP syntax.


ytdHuang (approved):
LGTM!

PR #247, opened 2024-10-01
[DOCS] Reduce figures size in Documentation

albertomercurio:
## Description
Here I reduce the size of the Figures in the Documentation, in order to have a better reading experience.


ytdHuang (approved):
LGTM!

PR #250, opened 2024-10-02
[DOCS] add Documentation about type instabilities

albertomercurio:
## Description

Here I write the page about type-instabilities. It is a specific page for people interested in performances on their numerical calculations.

### DIFF
--- a/docs/src/type_stability.md
+++ b/docs/src/type_stability.md
@@ -1,3 +1,219 @@
 # [The Importance of Type-Stability](@id doc:Type-Stability)
 
-This page is still under construction.
+You are here because you have probably heard about the excellent performance of Julia compared to other common programming languages like Python. One of the reasons is the Just-In-Time (JIT) compiler of Julia, which is able to generate highly optimized machine code. However, the JIT compiler can only do its job if the code type can be inferred. What does this mean exactly? Here we try to explain it in a simple way.
+
+!!! note
+    This page is not a tutorial on `QuantumToolbox.jl`, but rather a general guide to writing Julia code for simulating quantum systems efficiently. If you don't care about the performance of your code, you can skip this page.
+
+## Basics of type stability
+
+Let's have a look at the following example:
+
+```@setup type-stability
+using InteractiveUtils
+using QuantumToolbox
+```
+
+```@example type-stability
+function foo(x)
+    if x > 0
+        return 1
+    else
+        return -1.0
+    end
+end
+```
+
+The function `foo` apparently seems to be innocent. It takes an argument `x` and returns either `1` or `-1.0` depending on the sign of `x`. However, the return type of `foo` is not clear. If `x` is positive, the return type is `Int`, otherwise it is `Float64`. This is a problem for the JIT compiler, because it has to determine the return type of `foo` at runtime. This is called type instability (even though it is a weak form) and may lead to a significant performance penalty. To avoid this, always aim for type-stable code. This means that the return type of a function should be clear from the types of its arguments. We can check the inferred return type of `foo` using the `@code_warntype` macro:
+
+```@example type-stability
+@code_warntype foo(1)
+```
+
+The key point is to ensure the return type of a function is clear from the types of its arguments. There are several ways to achieve this, and the best approach depends on the specific problem. For example, one can use the same return type:
+
+```@example type-stability
+function foo(x)
+    if x > 0
+        return 1.0
+    else
+        return -1.0
+    end
+end
+```
+
+Or you can ensure the return type matches the type of the argument:
+
+```@example type-stability
+function foo(x::T) where T
+    if x > 0
+        return T(1)
+    else
+        return -T(1)
+    end
+end
+```
+
+The latter example is very important because it takes advantage of Julia's multiple dispatch, which is one of the most powerful features of the language. Depending on the type `T` of the argument `x`, the Julia compiler generates a specialized version of `foo` that is optimized for this type. If the input type is an `Int64`, the return type is `Int64`, if `x` is a `Float64`, the return type is `Float64`, and so on.
+
+```@example type-stability
+@show foo(1)
+@show foo(-4.4)
+@show foo(1//2)
+nothing # hide
+```
+
+!!! note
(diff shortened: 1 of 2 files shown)


docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
 # [The Importance of Type-Stability](@id doc:Type-Stability)
 
-This page is still under construction.
+You are here because you have probably heard about the excellent performance of Julia compared to other common programming languages like Python. One of the reasons is the Just-In-Time (JIT) compiler of Julia, which is able to generate highly optimized machine code. However, the JIT compiler can only do its job if the code type can be inferred. What does this mean exactly? Here we try to explain it in a simple way.
```
ytdHuang, line 3: I think it would be better to also mention julia official document page here.
like:

~~What does this mean exactly? Here we try to explain it in a simple way.~~
You can also read the section [Performance Tips](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-tips) in Julia's official documentation for more details.

Here, we try to explain it briefly.

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+
+```@example type-stability
+function foo(x)
+    if x > 0
+        return 1
+    else
+        return -1.0
+    end
+end
+```
```
ytdHuang, line 25: maybe add
```julia
nothing # hide
```
here

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+
+```@example type-stability
+function foo(x)
+    if x > 0
+        return 1.0
+    else
+        return -1.0
+    end
+end
+```
```
ytdHuang, line 43: maybe add
```julia
nothing # hide
```
here

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+
+```@example type-stability
+function foo(x::T) where T
+    if x > 0
+        return T(1)
+    else
+        return -T(1)
+    end
+end
+```
```
ytdHuang, line 55: maybe add
```julia
nothing # hide
```
here

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+y = 2.4
+
+function bar(x)
+    res = zero(x) # this returns the zero of the same type of x
+    for i in 1:1000
+        res += y * x
+    end
+    return res
+end
+```
```
ytdHuang, line 83: maybe add
```julia
nothing # hide
```
here

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+
+```@example type-stability
+σx_2.data
+```
+
+```@example type-stability
+σx_2.type
+```
+
+`Operator` is a synonym for [`OperatorQuantumObject`](@ref).
```
ytdHuang, line 173: maybe use [Operator](@ref) here

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+
+!!! note
+    This page is not a tutorial on `QuantumToolbox.jl`, but rather a general guide to writing Julia code for simulating quantum systems efficiently. If you don't care about the performance of your code, you can skip this page.
+
+## Basics of type stability
+
+Let's have a look at the following example:
+
+```@setup type-stability
+using InteractiveUtils
```
ytdHuang, line 13: Do we need this ?
I remove it, and I can still build the docs locally.

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+
+While in the last example of the `foo` function we got a weak form of type instability, returning a `Union{Int, Float64}`, in this case the return type of `bar` is `Any`, meaning that the compiler doesn't know anything about the return type. Thus, this function has nothing different from a dynamically typed language like Python. We can benchmark the performance of `bar` using the [BenchmarkTools.jl](https://github.com/JuliaCI/BenchmarkTools.jl) package:
+
+```@example type-stability
+using BenchmarkTools
+
+@benchmark bar(3.2)
+```
+
+Here we see a lot of memory allocations and low performances in general. To fix this, we can declare a `const` variable instead:
```
ytdHuang, line 99: maybe add `const` (constant) ?
Just in case someone from Python doesn't understand the keyword (`const`)

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+σx_2.type
+```
+
+`Operator` is a synonym for [`OperatorQuantumObject`](@ref).
+
+```@example type-stability
+σx_2.dims
+```
+
+The `dims` field contains the dimensions of the subsystems (in this case, three subsystems with dimension 2 each). We can see that the type of `dims` is `SVector` instead of `Vector`. As we mentioned before, this is very useful in functions like [`ptrace`](@ref). Let's do a simple example of reshaping an operator internally generated from some `dims` input:
```
ytdHuang, line 179: dimension \`2\`

docs/src/type_stability.md
```diff
@@ -1,3 +1,219 @@
...
+
+```@example type-stability
+@benchmark reshape_operator_data($[2, 2, 2])
+```
+
+```@example type-stability
+@benchmark reshape_operator_data($((2, 2, 2)))
+```
+
+Which is an innocuous but huge difference in terms of performance. Hence, we highly recommend using `Tuple` or `SVector` when defining the dimensions of a user-defined quantum object.
```
ytdHuang, line 218: maybe use [\`QuantumObject\`](@ref)

docs/src/type_stability.md
```diff
@@ -215,5 +219,67 @@ Finally, let's look at the benchmarks
 @benchmark reshape_operator_data($((2, 2, 2)))
 ```
 
-Which is an innocuous but huge difference in terms of performance. Hence, we highly recommend using `Tuple` or `SVector` when defining the dimensions of a user-defined quantum object.
+Which is an innocuous but huge difference in terms of performance. Hence, we highly recommend using `Tuple` or `SVector` when defining the dimensions of a user-defined [`QuantumObject`](@ref).
+
+### The use of `Val` in some `QuantumToolbox.jl` functions
```
ytdHuang, line 224: maybe use subsub-section `##` instead of subsubsub-section `###` ?

albertomercurio: Yes, I did a typo

ytdHuang (approved):

PR #251, opened 2024-10-03
Fix return type precision for `n_thermal`

ytdHuang:
As title

### DIFF
--- a/src/utilities.jl
+++ b/src/utilities.jl
@@ -42,10 +42,10 @@ n(\omega, \omega_{\textrm{th}}) = \frac{1}{e^{\omega/\omega_{\textrm{th}}} - 1},
 ```
 where ``\hbar`` is the reduced Planck constant, and ``k_B`` is the Boltzmann constant.
 """
-function n_thermal(ω::Real, ω_th::Real)::Float64
+function n_thermal(ω::T1, ω_th::T2) where {T1<:Real,T2<:Real}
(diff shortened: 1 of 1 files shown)


src/utilities.jl
```diff
@@ -42,10 +42,10 @@ n(\omega, \omega_{\textrm{th}}) = \frac{1}{e^{\omega/\omega_{\textrm{th}}} - 1},
 ```
 where ``\hbar`` is the reduced Planck constant, and ``k_B`` is the Boltzmann constant.
 """
-function n_thermal(ω::Real, ω_th::Real)::Float64
+function n_thermal(ω::T1, ω_th::T2) where {T1<:Real,T2<:Real}
     x = exp(ω / ω_th)
-    n = ((x != 1) && (ω_th > 0)) ? (1.0 / (x - 1.0)) : 0.0
-    return n
+    n = ((x != 1) && (ω_th > 0)) ? 1 / (x - 1) : 0
+    return _FType(promote_type(T1, T2))(n)
```
albertomercurio, line 48: What if the input types are for example `BigFloat`? Does the `_FType` know how to do?

ytdHuang: I think we need to add extended method for `_FType` in the future

albertomercurio: Ok, if we need just to extend `FType` we can merge it.

albertomercurio (approved):

PR #262, opened 2024-10-09
Add `Julia v1.11` to CI and buildkite config

ytdHuang:
Since the Julia v1.11 has released, we should update the CI config and buildkite config


albertomercurio:
What about the other CIs? I don't remember, like the benchmark tracking and others?

ytdHuang:
> What about the other CIs? I don't remember, like the benchmark tracking and others?

Those always use `version = 1`, so should be okay.

albertomercurio (approved):

PR #269, opened 2024-10-17
Clean low-rank mesolve and add Documentation bibliography

albertomercurio:
## Description
With this PR I clean the low rank master equation code, including the related documentation and the docstrings. Now everything should be much clearer.

I have also added the possibility to build a bibliography with bibtex, such that all the references of the documentation are seamlessly stored in one single page.

In the future I will further improve those functions, but for the moment this may be ok for a good documentation.

### DIFF
--- a/docs/src/tutorials/lowrank.md
+++ b/docs/src/tutorials/lowrank.md
@@ -1,5 +1,19 @@
 # [Low rank master equation](@id doc-tutor:Low-rank-master-equation)
 
+In this tutorial, we will show how to solve the master equation using the low-rank method. For a detailed explaination of the method, we recommend to read the article [gravina2024adaptive](@cite).
+
+As a test, we will consider the dissipative Ising model with a transverse field. The Hamiltonian is given by
+
+```math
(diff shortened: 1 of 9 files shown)


docs/src/tutorials/lowrank.md
```diff
@@ -21,42 +35,42 @@ Define lr-space dimensions
...
-z = hcat(broadcast(x -> x.data, ϕ)...)
-p0 = 0.0 # Population of the lr states other than the initial state
-B = Matrix(Diagonal([1 + 0im; p0 * ones(M - 1)]))
+z = hcat(get_data.(ϕ)...)
+B = Matrix(Diagonal([1 + 0im; zeros(M - 1)]))
 S = z' * z # Overlap matrix
 B = B / tr(S * B) # Normalize B
 
-ρ = QuantumObject(z * B * z', dims = ones(Int, N_modes) * N_cut); # Full density matrix
+ρ = QuantumObject(z * B * z', dims = ntuple(i->N_cut, Val(N_modes))); # Full density matrix
```
ytdHuang, line 73: I did a benchmark.
The following method for generating tuple-type `dims` is faster and less memory allocation:
```julia
Tuple(ones(Int, N_modes) * N_cut)
```

ytdHuang (approved):

ytdHuang:
Okay
I approved it to make sure GitHub has the record that I've gone throught it.
So that if there is new commits than it can tell the differences.

PR #277, opened 2024-10-30
Support time-dependent `liouvillian`

ytdHuang:
## Description
This PR support constructing time-dependent Liouvillian from a given `Operator` type `QobjEvo`.

## Additional context
@albertomercurio 
Do you think we can remove those `f::Function = identity (liouvillian)` keyword arguments now ?

### DIFF
--- a/src/time_evolution/mesolve.jl
+++ b/src/time_evolution/mesolve.jl
@@ -42,23 +42,7 @@ function _generate_mesolve_kwargs(e_ops, progress_bar::Val{false}, tlist, kwargs
 end
 
 _mesolve_make_L_QobjEvo(H::QuantumObject, c_ops) = QobjEvo(liouvillian(H, c_ops); type = SuperOperator)
-function _mesolve_make_L_QobjEvo(H::Tuple, c_ops)
-    c_ops isa Nothing && return QobjEvo(H; type = SuperOperator, f = liouvillian)
-    return QobjEvo((H..., mapreduce(op -> lindblad_dissipator(op), +, c_ops)); type = SuperOperator, f = liouvillian)
-end
-_mesolve_make_L_QobjEvo(H::QuantumObjectEvolution{DT,OperatorQuantumObject}, c_ops) where {DT<:AbstractSciMLOperator} =
-    throw(
-        ArgumentError(
-            "This function does not support the data type of time-dependent Operator `H` currently. Try to provide `H` as a time-dependent SuperOperator or Tuple instead.",
-        ),
-    )
-function _mesolve_make_L_QobjEvo(
-    H::QuantumObjectEvolution{DT,SuperOperatorQuantumObject},
-    c_ops,
-) where {DT<:AbstractSciMLOperator}
-    c_ops isa Nothing && return H
-    return H + QobjEvo((mapreduce(op -> lindblad_dissipator(op), +, c_ops)))
-end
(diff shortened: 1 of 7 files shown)


src/time_evolution/mesolve.jl
```diff
@@ -42,23 +42,7 @@ function _generate_mesolve_kwargs(e_ops, progress_bar::Val{false}, tlist, kwargs
...
-        ),
-    )
-function _mesolve_make_L_QobjEvo(
-    H::QuantumObjectEvolution{DT,SuperOperatorQuantumObject},
-    c_ops,
-) where {DT<:AbstractSciMLOperator}
-    c_ops isa Nothing && return H
-    return H + QobjEvo((mapreduce(op -> lindblad_dissipator(op), +, c_ops)))
-end
+_mesolve_make_L_QobjEvo(H::Union{QuantumObjectEvolution,Tuple}, c_ops) = liouvillian(QobjEvo(H), c_ops)
```
albertomercurio, line 45: Can a `liouvillian` between a `QobjEvo` and `c_ops` as a list of `Qobj` be performed? How?

test/core-test/quantum_objects_evo.jl
```diff
@@ -160,25 +160,37 @@
...
+        # Operator
+        H_td = QobjEvo(((a, coef1), a' * a, (a', coef2)))
+        H_ti = coef1(p, t) * a + a' * a + coef2(p, t) * a'
+        ψ = rand_ket(N)
+        @test H_td(p, t) ≈ H_ti
+        @test H_td(ψ, p, t) ≈ H_ti * ψ
         @test isconstant(a) == true
-        @test isconstant(op1) == false
+        @test isconstant(H_td) == false
         @test isconstant(Qobj(a)) == true
```
albertomercurio, line 179: Do we agree that this test should be `inconstant(QobjEvo(a)) == true`? Maybe I made a mistake in the previous PR.

src/time_evolution/mesolve.jl
ytdHuang: as I changed definition of `liouvillian` in this PR:
```julia
function liouvillian(
    H::AbstractQuantumObject{DT,OpType},
    c_ops::Union{Nothing,AbstractVector,Tuple} = nothing,
    Id_cache = I(prod(H.dims)),
) where {DT,OpType<:Union{OperatorQuantumObject,SuperOperatorQuantumObject}}

    # if H is Operator QbojEvo, it will create `-1im * (spre - spost)` for each AbstractSciMLOperator in H.data.ops
    L = liouvillian(H, Id_cache)
    if !(c_ops isa Nothing)
        # Here, it's just a QobjEvo (L) + Qobj (lindblad_dissipator), which will become QobjEvo)
        L += mapreduce(lindblad_dissipator, +, c_ops)
    end
    return L
end
```

test/core-test/quantum_objects_evo.jl
ytdHuang: ok
will update it

src/time_evolution/mesolve.jl
albertomercurio: Perhaps it is then better to first create something like `L2 = mapreduce(lindblad_dissipator, +, c_ops)` and then doing `L + L2`. In this way, we first create a simple `Qobj` (`L2`), and then we sum it to `L`. Otherwise, we will get an `AddedOperator` with many `ops` as the number of `c_ops`.

albertomercurio: Ok, `mapreduce` already does this. Never mind.

albertomercurio (approved):

PR #280, opened 2024-10-31
Support time-dependent `lindblad_dissipator`

ytdHuang:
## Description
As title

TODO:
- [x] support `lindblad_dissipator(::QobjEvo)`
- [x] prompt a warning and use lazy tensor in `_spre` and `_spost` for un-handled `AbstractSciMLOperator`
- [x] check whether `c_ops` including different `Qobj` and `QobjEvo` can work
- [x] fix type check in `QobjEvo`
- [x] move `get_typename_wrapper(A)` to `src/utilities.jl` and support for `A::Any` instead of `A::AbstractQuantumObject`
- [x] remove `f::Function = identity (liouvillian)`
- [x] `@reexport` some functions in `SciMLOperators.jl` for `AbstractQuantumObject` (e.g., `isconstant`, `iscached`, `cache_operator`, etc)

### DIFF
--- a/src/QuantumToolbox.jl
+++ b/src/QuantumToolbox.jl
@@ -8,16 +8,7 @@ import Reexport: @reexport
 @reexport import StaticArraysCore: SVector
 @reexport using LinearAlgebra
 @reexport using SparseArrays
-@reexport import SciMLOperators:
-    AbstractSciMLOperator,
-    MatrixOperator,
-    ScalarOperator,
-    ScaledOperator,
-    AddedOperator,
-    IdentityOperator,
-    cache_operator,
-    iscached,
-    isconstant
+@reexport import SciMLOperators: cache_operator, iscached, isconstant
 
 # other functions in LinearAlgebra
 import LinearAlgebra: BlasReal, BlasInt, BlasFloat, BlasComplex, checksquare
@@ -44,7 +35,16 @@ import SciMLBase:
     ContinuousCallback,
     DiscreteCallback
 import StochasticDiffEq: StochasticDiffEqAlgorithm, SRA1
-import SciMLOperators: SciMLOperators, update_coefficients!, concretize
+import SciMLOperators:
+    SciMLOperators,
+    AbstractSciMLOperator,
+    MatrixOperator,
+    ScalarOperator,
+    ScaledOperator,
+    AddedOperator,
+    IdentityOperator,
+    update_coefficients!,
+    concretize
 import LinearSolve: LinearProblem, SciMLLinearSolveAlgorithm, KrylovJL_MINRES, KrylovJL_GMRES
 import DiffEqBase: get_tstops
 import DiffEqCallbacks: PeriodicCallback, PresetTimeCallback, TerminateSteadyState
--- a/src/QuantumToolbox.jl
+++ b/src/QuantumToolbox.jl
@@ -3,17 +3,28 @@ module QuantumToolbox
 # Re-export:
 #   1. StaticArraysCore.SVector for the type of dims
 #   2. basic functions in LinearAlgebra and SparseArrays
+#   3. some functions in SciMLOperators
 import Reexport: @reexport
 @reexport import StaticArraysCore: SVector
 @reexport using LinearAlgebra
 @reexport using SparseArrays
+@reexport import SciMLOperators:
+    AbstractSciMLOperator,
+    MatrixOperator,
+    ScalarOperator,
+    ScaledOperator,
+    AddedOperator,
+    IdentityOperator,
+    cache_operator,
+    iscached,
+    isconstant
 
 # other functions in LinearAlgebra
 import LinearAlgebra: BlasReal, BlasInt, BlasFloat, BlasComplex, checksquare
 import LinearAlgebra.BLAS: @blasfunc
 import LinearAlgebra.LAPACK: hseqr!
 
-# SciML packages (for OrdinaryDiffEq and LinearSolve)
+# SciML packages (for QobjEvo, OrdinaryDiffEq, and LinearSolve)
 import SciMLBase:
     solve,
     solve!,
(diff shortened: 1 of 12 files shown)


src/QuantumToolbox.jl
```diff
@@ -3,17 +3,28 @@ module QuantumToolbox
...
 #   1. StaticArraysCore.SVector for the type of dims
 #   2. basic functions in LinearAlgebra and SparseArrays
+#   3. some functions in SciMLOperators
 import Reexport: @reexport
 @reexport import StaticArraysCore: SVector
 @reexport using LinearAlgebra
 @reexport using SparseArrays
+@reexport import SciMLOperators:
+    AbstractSciMLOperator,
+    MatrixOperator,
```
albertomercurio, line 13: Do we really need to re-export all these functions? maybe just `isconstant` is fine. People using them would probably require to do advanced stuff, where importing `SciMLOperators` would be better.

src/qobj/boolean_functions.jl
```diff
@@ -91,8 +91,15 @@ isunitary(U::QuantumObject{<:AbstractArray{T}}; kwargs...) where {T} =
     isoper(U) ? isapprox(U.data * U.data', I(size(U, 1)); kwargs...) : false
 
 @doc raw"""
-    isconstant(A::AbstractQuantumObject)
+    SciMLOperators.iscached(A::AbstractQuantumObject)
+
+Test whether the [`AbstractQuantumObject`](@ref) `A` has preallocated caches for inplace evaluations.
+"""
+SciMLOperators.iscached(A::AbstractQuantumObject) = iscached(A.data)
```
albertomercurio, line 98: Does `iscached` work for `AbstracMatrix`?

src/qobj/superoperators.jl
```diff
@@ -14,19 +14,51 @@ _sprepost(A::AbstractMatrix, B::AbstractMatrix) = kron(transpose(sparse(B)), spa
 _sprepost(A::AbstractMatrix, B::AbstractSparseMatrix) = kron(transpose(B), sparse(A))
 _sprepost(A::AbstractSparseMatrix, B::AbstractMatrix) = kron(transpose(sparse(B)), A)
 _sprepost(A::AbstractSparseMatrix, B::AbstractSparseMatrix) = kron(transpose(B), A)
-_liouvillian(H::AbstractMatrix, Id::AbstractMatrix) = -1im * (_spre(H, Id) - _spost(H, Id))
+function _sprepost(A, B) # for any other input types
```
albertomercurio, line 17: Do we use this method of the function?

src/utilities.jl
```diff
@@ -128,6 +128,8 @@ function convert_unit(value::T, unit1::Symbol, unit2::Symbol) where {T<:Real}
     return _FType(T)(value * (_energy_units[unit1] / _energy_units[unit2]))
 end
 
+get_typename_wrapper(A) = Base.typename(typeof(A)).wrapper
```
albertomercurio, line 131: Why should we support this for arbitrary input?

test/core-test/quantum_objects_evo.jl
```diff
@@ -165,32 +165,51 @@
...
         @test isoper(H_td) == true
 
         # SuperOperator
-        c_ops = [sqrt(rand()) * a]
-        L_td = liouvillian(H_td, c_ops)
-        L_td2 = -1im * spre(H_td) + 1im * spost(H_td) + lindblad_dissipator(c_ops[1])
+        X = a * a'
+        c_op1 = QobjEvo(((a', coef1),))
+        c_op2 = QobjEvo(((a, coef2), (X, coef3)))
+        c_ops = [c_op1, c_op2]
```
albertomercurio, line 190: The two `QobjEvo` have different types. Using `Vector` would introduce some type instabilities. Although it is still safe here in the test, I think that we should always use type stable codes. So far we safely used `Vector` because there was no need. But here we should require to use `Tuple` instead. Speaking of, do we have a check for type inference of the `liouvillian` and `lindblad_dissipator` functions?

src/QuantumToolbox.jl
ytdHuang: maybe
`isconstant`, `iscached`, and `cache_operator`

src/qobj/superoperators.jl
ytdHuang: yes, when A or B are un-handled AbstractSciMLIperator

src/utilities.jl
ytdHuang: for the function
`_lazy_tensor_warning`

test/core-test/quantum_objects_evo.jl
ytdHuang: I can do it tomorrow.

src/QuantumToolbox.jl
ytdHuang: Done

src/qobj/boolean_functions.jl
ytdHuang: Done

src/QuantumToolbox.jl
```diff
@@ -3,17 +3,19 @@ module QuantumToolbox
 # Re-export:
 #   1. StaticArraysCore.SVector for the type of dims
 #   2. basic functions in LinearAlgebra and SparseArrays
+#   3. some functions in SciMLOperators
 import Reexport: @reexport
 @reexport import StaticArraysCore: SVector
 @reexport using LinearAlgebra
 @reexport using SparseArrays
+@reexport import SciMLOperators: cache_operator, iscached, isconstant
```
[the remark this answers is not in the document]
ytdHuang: But as you can see in the test I added.
even with that simple case, we need to cache operator manually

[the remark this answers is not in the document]
albertomercurio: Ok just because we are using the Lazy tensor operator, which we should avoid in general. 

Let say that we reexport this, but we should avoid this as much as we can, because these functions then become as part of QuantumToolbox.jl instead of the parent package, which might lead to confusion.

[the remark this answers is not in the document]
ytdHuang: Yeah
the reason I implement this method is because. When people saw the un-cached error message from SciMLOperators. The message suggests people to do `L = cache_operator(L, u)`

albertomercurio (approved):

PR #283, opened 2024-11-02
Improve SciMLOperator handling of sesolveProblem

albertomercurio:
## Description

With this PR I improve the handling of SciMLOperator in `sesolveProblem`. More specifically, the `mcsolveProblem` was returning a `ScaledOperator` instead of a simpler `MatrixOperator`. Now, it returns a `MatrixOperator`, which should make the operations a bit faster.


ytdHuang (approved):
LGTM!

PR #284, opened 2024-11-04
Optimize precompilation during runtests

ytdHuang:
## Description
This PR reduce unnecessary precompilation during runtests.


Project.toml
```diff
@@ -33,6 +33,7 @@ CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 QuantumToolboxCUDAExt = "CUDA"
 
 [compat]
+Aqua = "0.8"
 ArrayInterface = "6, 7"
 CUDA = "5"
```
albertomercurio, line 38: Should we remove `CUDA` here?

albertomercurio (approved):

albertomercurio:
LGTM.

Project.toml
ytdHuang: We can't
we still have weak dependent on `CUDA`

PR #286, opened 2024-11-05
Add spell check CI and fix typos

ytdHuang:
## Description
As title.

"[typos](https://github.com/crate-ci/typos)" will check our spelling, tell us where each typo is located, and give us suggestions.

I also add the ignore file for the word: `ket`, since we use it a lot.

@albertomercurio 
If you agree with this, I will fix all the typos detected by this CI.

### DIFF
--- a/src/steadystate.jl
+++ b/src/steadystate.jl
@@ -106,11 +106,11 @@ function _steadystate(
     idx_range = collect(1:N)
     rows = _get_dense_similar(L_tmp, N)
     cols = _get_dense_similar(L_tmp, N)
-    datas = _get_dense_similar(L_tmp, N)
+    _data = _get_dense_similar(L_tmp, N)
     fill!(rows, 1)
     copyto!(cols, N .* (idx_range .- 1) .+ idx_range)
-    fill!(datas, weight)
-    Tn = sparse(rows, cols, datas, N^2, N^2)
+    fill!(_data, weight)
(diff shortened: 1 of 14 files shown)


albertomercurio:
Yes good idea. Do we correct all the spelling in this pr? Or we merge and we fix them in the future ones?

src/steadystate.jl
```diff
@@ -106,11 +106,11 @@ function _steadystate(
     idx_range = collect(1:N)
     rows = _get_dense_similar(L_tmp, N)
     cols = _get_dense_similar(L_tmp, N)
-    datas = _get_dense_similar(L_tmp, N)
+    _data = _get_dense_similar(L_tmp, N)
```
albertomercurio, line 109: I think it is better to leave `datas` and add it to the words to exclude. We should keep the code readable

ytdHuang: but `data` is an uncountable noun @@

ytdHuang: or something like `L_tmp2` ?

ytdHuang: This is just a temporary variable
So I prefer not adding it to the exclude list

albertomercurio: Ok, maybe `nzvals` or `vals`. What I don't like is that is is the only variable written with that underscore.

ytdHuang: ok, changed to `vals`

PR #287, opened 2024-11-08
Convert Documentation to Vitepress

albertomercurio:
## Description

This PR should already work. I don't know if there is anything else to do. It seems ok.


.github/workflows/documentation.yml
```diff
@@ -21,27 +21,39 @@ on:
...
-          julia --project=docs -e '
-            using Documenter: DocMeta, doctest
-            using QuantumToolbox
-            DocMeta.setdocmeta!(QuantumToolbox, :DocTestSetup, :(using QuantumToolbox); recursive=true)
-            doctest(QuantumToolbox)'
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # For authentication with GitHub Actions token
+          DOCUMENTER_KEY: ${{ secrets.DOCUMENTER_KEY }} # For authentication with SSH deploy key
+          GKSwstype: "100" # for Plots.jl plots (if you have them)
+          JULIA_DEBUG: "Documenter"
+          DATADEPS_ALWAYS_ACCEPT: true
```
ytdHuang, line 59: I think we don't need to change this part

.gitignore
```diff
@@ -4,7 +4,6 @@
 *.jl.cov
 *.jl.mem
 Manifest.toml
-docs/build/
 
 .vscode
 *.json
```
ytdHuang, line 9: I think you need to remove ignoreing `.json`

docs/make.jl
```diff
@@ -71,16 +72,28 @@ makedocs(;
...
+    #     prettyurls = get(ENV, "CI", "false") == "true",
+    #     canonical = "https://qutip.github.io/QuantumToolbox.jl",
+    #     edit_link = "main",
+    #     assets = ["assets/favicon.ico"],
+    #     mathengine = MathEngine,
+    #     size_threshold_ignore = ["api.md"],
+    # ),
+    format = DocumenterVitepress.MarkdownVitepress(
+        repo = "https://qutip.github.io/QuantumToolbox.jl",
+        # deploy_url = "https://qutip.org/QuantumToolbox.jl/",
```
ytdHuang, line 85: I think we need to uncomment this one

.github/workflows/documentation.yml
albertomercurio: They said to copy their yaml file. I think that the only difference is the doctest (which we are not doing) and some env variables, right?

docs/make.jl
albertomercurio: Yes, I'm unsure. But using Documenter.jl, we didn't specify the deploy_url, everything was pushing to GitHub, like here. So, I think that the link is done in some other place

PR #289, opened 2024-11-09
Add instructions about building docs locally

ytdHuang:
## Description
As title


albertomercurio:
Should we also update the Makefile? Do you want to do it in this PR or in another one?

PR #304, opened 2024-11-12
add allocations tests and clean time evolution tests

albertomercurio:
## Description

With this PR I add the test for the number of allocations in all the time-evolution functions, in order to keep track of the memory allocations in the future.

Moreover, I clean a bit the file. I defined a global problem, and from that I tested all the cases like `sesolve`, `mesolve`, ....

### DIFF
--- a/test/core-test/time_evolution.jl
+++ b/test/core-test/time_evolution.jl
@@ -14,7 +14,7 @@
     # Jaynes-Cummings Hamiltonian
     H = ωc * a' * a + ωq / 2 * σz + g * (a' * σm + a * σm')
     ψ0 = kron(fock(N, 0), fock(2, 0))
-    tlist = range(0, 1000, 1000)
+    
     e_ops = [a' * a, σz]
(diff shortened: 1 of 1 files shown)


test/core-test/time_evolution.jl
```diff
@@ -1,32 +1,48 @@
...
+    g = 0.01
+    ωc = 1
+    ωq = 0.99
+    γ = 0.1
+    nth = 0.001
+
+    # Jaynes-Cummings Hamiltonian
+    H = ωc * a' * a + ωq / 2 * σz + g * (a' * σm + a * σm')
+    ψ0 = kron(fock(N, 0), fock(2, 0))
+    tlist = range(0, 1000, 1000)
```
ytdHuang, line 17: Is it necessary to define `tlist` here ?

Cause I saw sometimes you define `tlist` below, sometimes not.

Or maybe just define `tlist` in each `testset` individually to avoid ambiguity?

ytdHuang:
Hmm, somehow the format check doesn't pass

ytdHuang (approved):

albertomercurio:
Now it should be fine

PR #310, opened 2024-11-14
Try to fix several missing favicons in doc

ytdHuang:
## Description
I don't know why the favicon sometimes show up, sometimes disappeared

I just upload a `favicon.ico` file into the root directory in `gh-pages`, and add an extra link for it in `config.mts`.


albertomercurio:
But what if we use the `assets` kwarg in the [public API](https://luxdl.github.io/DocumenterVitepress.jl/v0.1.3/api#DocumenterVitepress.MarkdownVitepress) instead of modifying the raw code anytime?

albertomercurio (approved):

PR #314, opened 2024-11-19
Add Benchmark Comparison

albertomercurio:
## Description

As title, I added a benchmark comparison between QuantumToolbox.jl, QuTiP, dynamiqs, and QuantumOptics.jl. I made a script file which makes everything reproducible. I added the plot on both the Documentation and the README.


.gitignore
```diff
@@ -3,7 +3,7 @@
 *.jl.*.cov
 *.jl.cov
 *.jl.mem
-Manifest.toml
+/Manifest.toml
```
ytdHuang, line 6: I think this need to be rebased.

benchmarks/.gitignore
```diff
@@ -0,0 +1 @@
+/Manifest.toml
```
ytdHuang, line 1: I think we don't need this file

docs/src/resources/package_comparison.md
ytdHuang, line 1: This file should not be tracked by git

benchmarks/package_comparison/Manifest.toml
ytdHuang, line 1: This file should be ignored

ytdHuang:
The figure should not be tracked by git, cause it keeps updating everytime the script is executed.
So, I suggest we don't put the figure in `README.md`

.gitignore
albertomercurio: No, I need it to keep the Manifest of the benchmark comparison, otherwise it removes every Manifest file. I need the Manifest because I want reproducibility, like also the requirements.txt for the python packages.

benchmarks/.gitignore
albertomercurio: Same as above

benchmarks/package_comparison/Manifest.toml
albertomercurio: Sa me as above

docs/src/resources/package_comparison.md
albertomercurio: This file is executed only locally, to have the best performances in the comparison. So it would be just a static file

albertomercurio: It will change only when we re-run the benchmarks comparison locally.

PR #330, opened 2024-12-02
Make `spectrum` and `correlation` align with QuTiP

ytdHuang:
## Checklist

- [x] Please read [Contributing to QuantumToolbox.jl](https://qutip.org/QuantumToolbox.jl/stable/resources/contributing).

## Description
The current implementation of `spectrum` does not align with QuTiP (see the documentation [here](https://qutip.readthedocs.io/en/qutip-5.0.x/guide/guide-correlation.html)). The `spectrum` should only return the value of spectrum corresponding to the user specified `ω_list`. Also, the peudo-inverse method is missing.

Furthermore, in QuTiP, the spectra obtained from `fft` of the correlation function is done by a function called `spectrum_correlation_fft` instead of `spectrum`.

Finally, we can support `nothing` input state for correlation functions, that will automatically solve the `steadystate` first.

TODO list:

- [x] separate the source code of `spectrum` and `correlation` functions into independent files
- [x] adjust the parameter order of `spectrum` and `correlation` functions to align with `QuTiP` (also add the deprecated methods to `deprecated.jl`)
- [x] add time list checks for `correlation` functions
- [x] adjust the return correlation `Matrix` dimension order to align with `QuTiP`
- [x] remove the return of `ω_list` in `spectrum` function
- [x] support `nothing::Nothing` input state for correlation functions [only when Hamiltonian is constant (`Qobj`)]
- [x] allow time-dependent Hamiltonian (`QobjEvo`) as an input for `correlation` functions.
- [x] remove spectrum solver `FFTCorrelation` (move it to `deprecated.jl`)
- [x] introduce a new function called `spectrum_correlation_fft(tlist, corr)` to replace the functionality of `FFTCorrelation`
- [x] introduce `PseudoInverse <: SpectrumSolver` for `spectrum`
- [x] introduce `correlation_3op_1t`
- [x] write the document page: `Two-time correlation functions`

### DIFF
--- a/src/correlations.jl
+++ b/src/correlations.jl
@@ -39,7 +39,8 @@ function correlation_3op_2t(
         ψ0 = steadystate(L)
     end
 
-    allequal((L.dims, ψ0.dims, A.dims, B.dims, C.dims)) || throw(DimensionMismatch("The quantum objects are not of the same Hilbert dimension."))
+    allequal((L.dims, ψ0.dims, A.dims, B.dims, C.dims)) ||
+        throw(DimensionMismatch("The quantum objects are not of the same Hilbert dimension."))
 
     kwargs2 = merge((saveat = collect(tlist),), (; kwargs...))
     ρt = mesolve(L, ψ0, tlist; kwargs2...).states
--- a/src/correlations.jl
+++ b/src/correlations.jl
@@ -1,8 +1,10 @@
 export correlation_3op_2t, correlation_3op_1t, correlation_2op_2t, correlation_2op_1t
 
 function _check_correlation_time_list(tlist::AbstractVector)
-    any(t -> t == 0, tlist) || throw(ArgumentError("The time list for calculating correlation function must contain the element `0`"))
-    all(>=(0), tlist) || throw(ArgumentError("All the elements in the time list for calculating correlation function must be positive."))
+    any(t -> t == 0, tlist) ||
+        throw(ArgumentError("The time list for calculating correlation function must contain the element `0`"))
+    all(>=(0), tlist) ||
+        throw(ArgumentError("All the elements in the time list for calculating correlation function must be positive."))
     return nothing
 end
 
--- a/src/correlations.jl
+++ b/src/correlations.jl
@@ -55,9 +55,9 @@ function correlation_3op_2t(
         throw(DimensionMismatch("The quantum objects are not of the same Hilbert dimension."))
 
     kwargs2 = merge((saveat = collect(tlist),), (; kwargs...))
-    ρt = mesolve(L, ψ0, tlist; kwargs2...).states
+    ρt_list = mesolve(L, ψ0, tlist; kwargs2...).states
 
-    corr = map((t, ρ) -> mesolve(L, C * ρ * A, τlist .+ t, e_ops = [B]; kwargs...).expect[1, :], tlist, ρt)
+    corr = map((t, ρt) -> mesolve(L, C * ρt * A, τlist .+ t, e_ops = [B]; kwargs...).expect[1, :], tlist, ρt_list)
 
     # make the output correlation Matrix align with QuTiP
     # 1st dimension corresponds to tlist
--- a/src/correlations.jl
+++ b/src/correlations.jl
@@ -21,7 +21,7 @@ end
 
 Returns the two-times correlation function of three operators ``\hat{A}``, ``\hat{B}`` and ``\hat{C}``: ``\left\langle \hat{A}(t) \hat{B}(t + \tau) \hat{C}(t) \right\rangle`` for a given initial state ``|\psi_0\rangle``.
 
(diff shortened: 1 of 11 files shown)


albertomercurio:
Great

albertomercurio:
I would also change the order of the arguments, since I think it is a bit different from QuTiP.

ytdHuang:
> I would also change the order of the arguments, since I think it is a bit different from QuTiP.

I also did it in this PR~

src/correlations.jl
```diff
@@ -1,246 +1,186 @@
...
 @doc raw"""
-    correlation_3op_2t(H::QuantumObject,
-        ψ0::QuantumObject,
-        t_l::AbstractVector,
-        τ_l::AbstractVector,
+    correlation_3op_2t(H::AbstractQuantumObject,
+        ψ0::Union{Nothing,QuantumObject},
+        tlist::AbstractVector,
+        τlist::AbstractVector,
+        c_ops::Union{Nothing,AbstractVector,Tuple},
```
albertomercurio, line 16: Here we force the users to insert the collapse operators. I think that this should left free, as one can in principle calculate the correlation function also for the closed system case. In think that qutip also supports this.

src/correlations.jl
```diff
@@ -1,246 +1,186 @@
...
-    τ_l::AbstractVector,
-    A::QuantumObject{<:AbstractArray{T3},OperatorQuantumObject},
-    B::QuantumObject{<:AbstractArray{T4},OperatorQuantumObject},
-    C::QuantumObject{<:AbstractArray{T5},OperatorQuantumObject},
-    c_ops::Union{Nothing,AbstractVector,Tuple} = nothing;
+    H::AbstractQuantumObject{DataType,HOpType},
+    ψ0::Union{Nothing,QuantumObject{<:AbstractArray{T1},StateOpType}},
+    tlist::AbstractVector,
+    τlist::AbstractVector,
+    c_ops::Union{Nothing,AbstractVector,Tuple},
```
albertomercurio, line 31: Also here and in all the other cases below.

src/spectrum.jl
```diff
@@ -0,0 +1,177 @@
...
+    kwargs...,
+) where {MT1<:AbstractMatrix,T2,T3,HOpType<:Union{OperatorQuantumObject,SuperOperatorQuantumObject}}
+    return _spectrum(liouvillian(H, c_ops), ωlist, A, B, solver; kwargs...)
+end
+
+function _spectrum_get_rates_vecs_ss(L, solver::ExponentialSeries{T,true}) where {T}
+    result = eigen(L)
+    rates, vecs = result.values, result.vectors
+
+    return rates, vecs, steadystate(L).data
```
albertomercurio, line 64: I don't think we need this. The steady state is already computer in `eigen(L)`, so this is unnecessary I think.

src/spectrum.jl
```diff
@@ -0,0 +1,177 @@
...
+```
+
+See also the following list for `SpectrumSolver` docstrings:
+- [`ExponentialSeries`](@ref)
+- [`PseudoInverse`](@ref)
+"""
+function spectrum(
+    H::QuantumObject{MT1,HOpType},
+    ωlist::AbstractVector,
+    c_ops::Union{Nothing,AbstractVector,Tuple},
```
albertomercurio, line 51: Here instead, we need the collapse operators, I think. Because the spectrum needs some dissipation I would say.

src/correlations.jl
ytdHuang: just specify `nothing` or empty `c_ops` list.

QuTiP can only specify empty `c_ops` for this situation.

src/correlations.jl
ytdHuang: just specify `nothing` or empty `c_ops` list.

QuTiP can only specify empty `c_ops` for this situation.

src/spectrum.jl
ytdHuang: But we should still support `c_ops::Nothing`, cause users might directly specify `H` as Liouvillian.

src/spectrum.jl
ytdHuang: I just copied this part from the original file. lol

The `ExponentialSeries` provides two methods depend on the parameter `calc_steadystate = false`.

I think this part is for `calc_steadystate = true` ?

ytdHuang: from the data of `codecov`, our runtests didn't cover this part.

What do you think, do you want to keep this functionality ?

src/correlations.jl
albertomercurio: Okok fine

src/spectrum.jl
albertomercurio: Aaaah sorry! Now I remember we we do need this. Sometimes the steady state estimation is better when calling `steadystate` rather that diagonalizing the liouvillian

src/spectrum.jl
albertomercurio: Yeye

albertomercurio (approved):

PR #345, opened 2024-12-09
[Doc] Remove unnecessary URLs in bibtex

ytdHuang:
## Description
As title


docs/src/tutorials/lowrank.md
```diff
@@ -1,6 +1,6 @@
 # [Low rank master equation](@id doc-tutor:Low-rank-master-equation)
 
-In this tutorial, we will show how to solve the master equation using the low-rank method. For a detailed explanation of the method, we recommend to read the article [gravina2024adaptive](@cite).
+In this tutorial, we will show how to solve the master equation using the low-rank method. For a detailed explanation of the method, we recommend to read the article [gravina2024adaptive](@citet).
```
albertomercurio, line 3: Why do also need this?

ytdHuang: not necessary
but I think it is a better display for these cases

you can take a look at the DocumenterCitations documentation

albertomercurio: It seems [from here](https://luxdl.github.io/DocumenterVitepress.jl/v0.1.3/citations#author_year_style) that the two cases are equivalent in the case of author-year format. I would use the simple command `@cite` then.

albertomercurio: Ok, the `@cite` introduces some parenthesis, while the `@citet` not. Again, I would prefer the `@cite` case then

ytdHuang: umm
its different

(Author, year)
and
Author (year)

albertomercurio: Exactly, I think that it is better to keep the `(Author, year)` format. The parentheses help to separate the text from the citation

albertomercurio (approved):

PR #358, opened 2024-12-22
Add Zulip sponsor

albertomercurio:
Zulip has recently approved our request for a free upgrade to the Zulip Cloud Standard subscription. As requested, we are encouraged to acknowledge their sponsorship on our main page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,3 +176,18 @@ Here we provide a brief performance comparison between `QuantumToolbox.jl` and o
 You are most welcome to contribute to `QuantumToolbox.jl` development by forking this repository and sending pull requests (PRs), or filing bug reports at the issues page. You can also help out with users' questions, or discuss proposed changes in the [QuTiP discussion group](https://groups.google.com/g/qutip).
 
 For more information about contribution, including technical advice, please see the [Contributing to QuantumToolbox.jl](https://qutip.org/QuantumToolbox.jl/stable/resources/contributing) section of the documentation.
+
+## Acknowledgements
+
+`QuantumToolbox.jl` is supported by the [Unitary Fund](https://unitary.fund), a grant program for quantum technology projects. We are also grateful to the [Zulip](https://zulip.com) team for providing a free chat service for open-source projects.
+
+<!-- Acknowledgement Logos -->
+<div align="center">
+  <a href="https://unitary.fund">
+    <img src="https://raw.githubusercontent.com/unitaryfund/unitary.fund/refs/heads/main/src/assets/svg/logo.svg" alt="Unitary Fund logo" width="200">
+  </a>
+  &nbsp;&nbsp;&nbsp;&nbsp;
+  <a href="https://zulip.com">
+    <img src="https://zulip.com/static/images/logo/zulip-org-logo.svg" alt="Zulip logo" width="200">
+  </a>
+</div>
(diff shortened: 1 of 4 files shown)


ytdHuang:
Should we do it similarly like qutip (in their [README: Support](https://github.com/qutip/qutip) and also [documentation: Funding](https://qutip.readthedocs.io/en/qutip-5.0.x/frontmatter.html#funding))?

We might also need to add RIKEN and something else in the future.

albertomercurio:
Zulip doesn't exactly belong to the Funding class. They just gave us the upgraded cloud service and they simply asked to put their logo in our website. For the readme, both Support or Acknowledgements are fine to me.
 For the documentation, what about `Resources/Acknowledgements` or `Resources/Support`?

I would probably prefer Acknowledgements, since Support is too generic.

ytdHuang:
> Zulip doesn't exactly belong to the Funding class. They just gave us the upgraded cloud service and they simply asked to put their logo in our website. For the readme, both Support or Acknowledgements are fine to me. For the documentation, what about `Resources/Acknowledgements` or `Resources/Support`?
> 
> I would probably prefer Acknowledgements, since Support is too generic.

Maybe generic is better? We can add a few words (just like the way you did) to point out what kind of support does each of them gave us.

Cause in the future we might be supported by other organizations and grants (hope so) lol

So I suggest we call it Support for both README and documentation (`Resources/Support` is okay for me).

albertomercurio:
What I don't like of a "Support" section is it is too generic, as people may think it is related to how to support QuantumToolbox.jl, while "Acknowledgements" is already clear that we want to acknowledge fundings and other things there.

ytdHuang:
How about we call the section in Readme as `Acknowledgements`, and page in document as `resouces/acknowledgements`.

In Readme, we probably just need to put the logo there.

In the documentation, we can add sub-sections, i.e., `Funding`, `Support`, `Others`, or something else, to acknowledge different kinds of support individually?

README.md
```diff
@@ -176,3 +176,24 @@ Here we provide a brief performance comparison between `QuantumToolbox.jl` and o
...
+
+### Other Acknowledgements
+
+We are also grateful to the [Zulip](https://zulip.com) team for providing a free chat service for open-source projects.
+
+<!-- Acknowledgement Logos -->
+<div align="center">
+  <a href="https://unitary.fund">
+    <img src="https://raw.githubusercontent.com/unitaryfund/unitary.fund/refs/heads/main/src/assets/svg/logo.svg" alt="Unitary Fund logo" width="200">
+  </a>
```
ytdHuang, line 194: should we put this together in the `Fundings` section ?

docs/src/resources/acknowledgements.md
```diff
@@ -0,0 +1,20 @@
...
+
+## [Other Acknowledgements](@id doc:Other-Acknowledgements)
+
+We are also grateful to the [Zulip](https://zulip.com) team for providing a free chat service for open-source projects.
+
+<!-- Acknowledgement Logos -->
+<div align="center">
+  <a href="https://unitary.fund">
+    <img src="https://raw.githubusercontent.com/unitaryfund/unitary.fund/refs/heads/main/src/assets/svg/logo.svg" alt="Unitary Fund logo" width="200">
+  </a>
```
ytdHuang, line 15: same here

ytdHuang (approved):

README.md
albertomercurio: Ok, done.

PR #394, opened 2025-02-10
Change MultiSiteOperator function name to multisite_operator

albertomercurio:
## Description

I change the name of the `MultiSiteOperator` function to `multisite_operator`, since functions are always lower case here. Only structs can have upper case names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@ Release date: 2024-12-06
 ## [v0.23.0]
 Release date: 2024-12-04
 
-- Change `SingleSiteOperator` with the more general `MultiSiteOperator`. ([#324])
+- Change `SingleSiteOperator` with the more general `multisite_operator`. ([#324])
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -53,7 +54,7 @@ Release date: 2024-12-06
 ## [v0.23.0]
 Release date: 2024-12-04
 
-- Change `SingleSiteOperator` with the more general `MultiSiteOperator`. ([#324])
+- Change `SingleSiteOperator` with the more general `multisite_operator`. ([#324])
```
ytdHuang, line 57: We should not change the previous `ChangeLog`

ytdHuang (commented):
Would it be even easier to just name it as `multisite` (without `_operator`)?

albertomercurio:
It can be misleading as ve could also have in the future the equivalent of multisite state

ytdHuang:
> It can be misleading as ve could also have in the future the equivalent of multisite state

ok, I see

CHANGELOG.md
albertomercurio: Oh, true

ytdHuang (approved):

PR #412, opened 2025-02-18
Introduce `groundstate`

ytdHuang:
## Checklist

## Description
This PR introduce a new function: `groundstate` to calculate the ground state eigenvalue and corresponding eigenvector.

This function currently doesn't support `sparse` eigsolve. Because it seems that `sparse` eigsolve doesn't find the smallest eigenvalue first.

### DIFF
--- a/src/qobj/eigsolve.jl
+++ b/src/qobj/eigsolve.jl
@@ -3,7 +3,7 @@ Eigen solvers and results for QuantumObject
 =#
 
 export EigsolveResult
-export eigenenergies, eigenstates, eigsolve
+export eigenenergies, eigenstates, groundstate, eigsolve
 export eigsolve_al
 
 @doc raw"""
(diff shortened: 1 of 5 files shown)


src/qobj/eigsolve.jl
```diff
@@ -538,3 +544,40 @@ function eigenstates(
...
+    #    eigvals = safe ? 2 : 1  # number of eigenvalues to calculate
+    #    result = eigsolve(A; eigvals = eigvals, tol = tol, kwargs...)
+    #end
+
+    # the tolarence should be less strick than the `tol` for the eigensolver
+    # so it's numerical errors are not seen as degenerate states.
+    evals = result.values
+    safe && (abs(evals[2] - evals[1]) < (10 * tol)) && @warn "Ground state may be degenerate."
+
+    return evals[1], QuantumObject(result.vectors[:, 1], _eigenvector_type(OpType), result.dimensions)
```
albertomercurio, line 582: I would suggest to use `eigenstates` instead of `eigen`. In this way we can use all the internal `kwargs`, including also the `sparse` one.

In the case of `sparse`, it automatically finds the largest eigenvalue by the nature of the Arnoldi algorithm. However, this can be fixed by defining a `sigma` argument, which performs the shift-inverse procedure.

ytdHuang: That is why I didn't use the `eigenstates`
Because this function is only calculating for the ground state. We should assign some of the keyword arguments, make it not adjustable for the users.

ytdHuang: What value should I set for `sigma` in `eigsolve` to make it search for the lowest eigenvalue first?

albertomercurio: > Because this function is only for calculating ground state. We should assign some of the keyword arguments, make it not adjustable from the users.

What do you mean exactly? If we use `eigenstates(A; kwargs...)` we can forward all the `kwargs` to that, even the `sparse = ...` (which by default is `false`, so that we fall in this case).  If the user specify `sparse=true`, then he/she is supposed to be more advanced, and should know that the shift-inverse method applies here. By doing so, they should also specify a `sigma` shift value. We can even write a note on that.



> What value should I set for `sigma` in `eigsolve` to make it search for the lowest eigenvalue first?

`sigma` should be a value smaller than the supposed ground state energy. Let say that you expect the ground state energy to me close to 0, then `sigma=-0.1` is fine.

ytdHuang: So, in general, I should set `sigma = -Inf` ?

albertomercurio: No this won't work. This is problem dependent. 

The user using `sparse=true` is supposed to be advanced to be able to set this value.

ytdHuang: Hmm
Ok
I would probably close this PR.

Seems like we cannot make it.

PR #415, opened 2025-02-18
Differentiate tolerances between pure and mixed state dynamics

albertomercurio:
## Checklist

- [ ] Please read [Contributing to Quantum Toolbox in Julia](https://qutip.org/QuantumToolbox.jl/stable/resources/contributing).
- [ ] Any code changes were done in a way that does not break public API.

## Description
For the `sesolve`, `mcsolve`, and `mesolve` functions, we were using the tolerances given in QuTiP. However, QuTiP uses `adams` as default algorithm, which has a different behavior than `Tsit5`.

Here I slightly change the structure of the default ODE and SDE options tolerances. I make a dinstinction between pure and mixed states dynamics. Indeed, pure states dynamics generally require smaller tolerances, but this is not needed for density matrices evolutions.

This would allow to have better performances in `mesolve`, still having an accurate evolution.


ytdHuang:
Hmm

but this looks a bit weird for the users ?

albertomercurio:
As soon as the solver is different, having different tolerances is fine. Even with the same tolerances we can have different results. Moreover, in the future I'm planning to implement the `AutoAbsTol` callback, which dynamically changes the `abstol` depending on the dynamics. So their behavior is different anyhow.

albertomercurio:
I'm closing it as it is better to keep the tolerances as simple as possible.

PR #449, opened 2025-04-15
Add autodiff tests to GROUP All

albertomercurio:
## Description
Autodiff tests should always run together with the other tests, as all changes should always keep autodiff working.

So, I added the autodiff tests to the GROUP All


ytdHuang:
Not sure if we should do it...

It's gonna take very long time for local build...

Maybe just keep it the same as `Makie` extension ?

albertomercurio:
Okok

PR #456, opened 2025-04-29
Unify structure of `QuantumObjectType`

albertomercurio:
## Checklist

## Description
The current version uses both `OperatorQuantumObject` and `Operator` to handle the same type of `QuantumObject`. This can be a bit misleading. The same happens for all the other `QuantumObjectType`s (e.g., `Ket` or `SuperOperator`).

As an example of the ambiguities given by this, in the documentation we often write something like

> ... where `A` is a `Operator`

but code-wise `A` is a `OperatorQuantumObject`.

Here I try to unify everything, leaving only the shorter synonyms like `Operator`.

We only ask the users to call `QuantumObject(data, Operator(), dims)` instead of `QuantumObject(data, Operator, dims)`.


src/qobj/quantum_object_base.jl
```diff
@@ -28,95 +21,55 @@ abstract type AbstractQuantumObject{ObjType,DimType,DataType} end
...
 @doc raw"""
-    const Bra = BraQuantumObject()
+    Bra <: QuantumObjectType
 
-A constant representing the type of [`BraQuantumObject`](@ref): a bra state ``\langle\psi|``
+Constructor representing a bra state ``\langle\psi|``.
 """
-const Bra = BraQuantumObject()
+struct Bra <: QuantumObjectType end
+Base.show(io::IO, ::Bra) = print(io, "Bra")
```
ytdHuang, line 32: I think these overwriting of `Base.show` can be removed for `QuantumObjectType`

cause the default `show` and `print` of a structure is already correct.

like:

```julia
> struct Ket end

> k = Ket()
> print("type=$k")
type=Ket()
```

Therefore, printing an `Qobj/QobjEvo` should already show `type=Operator()` (with `()`) by default.

So that we don't need to change the `Base.show` for `Qobj/QobjEvo`

Furthermore, I think there should be some change in the runtests regarding to `Base.show`, like adding `()`

src/qobj/boolean_functions.jl
```diff
@@ -8,55 +8,55 @@ export isunitary
...
 
 @doc raw"""
     issuper(A)
 
-Checks if the [`AbstractQuantumObject`](@ref) `A` is a [`SuperOperatorQuantumObject`](@ref). Default case returns `false` for any other inputs.
+Checks if the [`AbstractQuantumObject`](@ref) `A` is a [`SuperOperator`](@ref). Default case returns `false` for any other inputs.
 """
 issuper(A::AbstractQuantumObject) = issuper(typeof(A))
-issuper(::Type{<:AbstractQuantumObject{SuperOperatorQuantumObject,N}}) where {N} = true
+issuper(::Type{<:AbstractQuantumObject{SuperOperator,N}}) where {N} = true
```
ytdHuang, line 59: Although we are not implementing other `SuperOperatorType` yet, but I think for `issuper`, we can already put:

```suggestion
issuper(::Type{<:AbstractQuantumObject{<:SuperOperatorType,N}}) where {N} = true
```

src/qobj/quantum_object_base.jl
albertomercurio: Ok

ytdHuang (approved):

PR #472, opened 2025-05-28
Implement Bloch Sphere rendering

### LINKED ISSUE #461: Implement Bloch Sphere rendering
### Problem Description

Implement the Bloch Sphere rendering. It should follow the same syntax as QuTiP, as showed in [this page](https://qutip.readthedocs.io/en/qutip-5.1.x/guide/guide-bloch.html).

It should also follow the methods already defined for the [`plot_wigner`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.plot_wigner) and [`plot_fock_distribution`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.plot_fock_distribution) functions, allowing for future possible alternative backends. `Makie` should be enough for the moment.

### Proposed Solution

The Bloch sphere constructor should be defined like

```julia
b = QuantumToolbox.Bloch()
```

All the attributes have to be defined using functions like

```julia
pnt = [1/sqrt(3), 1/sqrt(3), 1/sqrt(3)]
add_points!(pnt) # Notice the exclamation mark, as it changes the structure of the object
render(b)
```

### Additional Context

Makie Documentation: https://docs.makie.org

Fe-r-oz:
## Checklist

- [x] All documents (in `docs/` folder) related to code changes were updated and able to build locally by running: `make docs`. PS. make docs crashes my PC after it's ran for some time but I made sure to add the necessary documentation. 

## Description

Edit: This PR fixes #461. The goal is to add the bloch rendering in Julia which have a similar user experience as qutip

## Additional context

**Edit**: Attaching the plots that reproduce the entire [qutip documentation](https://qutip.readthedocs.io/en/stable/guide/guide-bloch.html) documentation:

### Add a Single Data Point

display

### Add a Single Vector

display

display

### Add Arc, Line, and Vector

display

### Add Multiple Points

display

display

display

display

We can also use `GLMakie`

Screenshot_select-area_20250604203219


albertomercurio:
Thank you, just a few other things before checking the code.

The sphere seems to be rotated compared to the one of QuTiP. Indeed, in QuTiP we have that x points almost outside, while here it is -y doing so. Better to align to QuTiP.

Since you added the `bloch.jl` file in the ext folder, it is better to embed everything inside a QuantumToolboxMakieExt folder.

The vectors tips seem to go outside the sphere.

Better to show all the examples showed in the [qutip documentation](https://qutip.readthedocs.io/en/stable/guide/guide-bloch.html), in order to check all the properties.

Fe-r-oz:
Some basic plots from qutip documentation. I will polish the rendering so the plots looks much better and accurate.

Figure 1 qutip:
display

Figure 2 qutip:
display

Figure 3 qutip:
display

Figure 4 qutip
display

Figure 5 qutip
display

docs/src/users_guide/extensions/cairomakie.md
```diff
@@ -19,4 +19,5 @@ The supported plotting functions are listed as follows:
 | **Plotting Function** | **Description** |
 |:----------------------|:----------------|
 | [`plot_wigner`](@ref) | [Wigner quasipropability distribution](https://en.wikipedia.org/wiki/Wigner_quasiprobability_distribution) |
-| [`plot_fock_distribution`](@ref) | [Fock state](https://en.wikipedia.org/wiki/Fock_state) distribution |
\ No newline at end of file
+| [`plot_fock_distribution`](@ref) | [Fock state](https://en.wikipedia.org/wiki/Fock_state) distribution |
+| [`plot_bloch`](@ref)  | [Qutip's Plotting on the Bloch Sphere](https://qutip.readthedocs.io/en/stable/guide/guide-bloch.html) |
```
albertomercurio: I would remove Qutip. Just "Plotting on the Bloch Sphere"

docs/src/users_guide/plotting_the_bloch_sphere.md
```diff
@@ -0,0 +1,103 @@
+# [Plotting on the Bloch Sphere](@id doc: plotting_the_bloch_sphere.md)
+
+```@setup Bloch_sphere_rendering
+using QuantumToolbox
+
+using Makie
```
albertomercurio: Your don't need to import also Makie. CairoMakie is enough.

docs/src/users_guide/plotting_the_bloch_sphere.md
```diff
@@ -0,0 +1,103 @@
...
+using CairoMakie
+CairoMakie.enable_only_mime!(MIME"image/svg+xml"())
+```
+
+## [Introduction](@id doc:Bloch_sphere_rendering)
+
+When studying the dynamics of a two-level system, it's often convenient to visualize the state of the system by plotting the state vector or density matrix on the Bloch sphere.
+
+In [QuantumToolbox.jl](https://qutip.org/QuantumToolbox.jl/), this can be done using the [`Bloch`](@ref) or [`plot_bloch`](@ref) methods that provide same syntax as [QuTiP](https://github.com/qutip/qutip).
```
albertomercurio: Maybe directly link to the Bloch Sphere page in qutip

https://qutip.readthedocs.io/en/stable/guide/guide-bloch.html

ext/QuantumToolboxMakieExt.jl
```diff
@@ -1,6 +1,8 @@
 module QuantumToolboxMakieExt
 
 using QuantumToolbox
+using LinearAlgebra
+using Makie
```
albertomercurio: Better to import only the necessary functions, if they are not a lot.

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +293,354 @@ raw"""
...
+b = Bloch()
+ψ₁ = normalize(basis(2, 0) + basis(2, 1))
+ψ₂ = normalize(basis(2, 0) - im * basis(2, 1))
+add_line!(b, ψ₁, ψ₂; fmt = "r--")
+```
+"""
+function QuantumToolbox.add_line!(
+    b::QuantumToolbox.Bloch,
+    start::QuantumObject{<:Union{Ket,Bra,Operator}},
+    endp::QuantumObject{<:Union{Ket,Bra,Operator}};
```
albertomercurio: `endp` is not a nice name. Better `end` or change both to `start_point` and `end_point`.

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +293,354 @@ raw"""
...
+        _state_to_bloch(start)
+    else
+        _dm_to_bloch(start)
+    end
+    p2 = if isket(endp) || isbra(endp)
+        _state_to_bloch(endp)
+    else
+        _dm_to_bloch(endp)
+    end
+    return add_line!(b, p1, p2; fmt = fmt, kwargs...)
```
albertomercurio: Why are we returning again `add_line!` with the same arguments? This may be misleading.

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +293,354 @@ raw"""
...
+function QuantumToolbox.add_states!(b::Bloch, states::Vector{<:QuantumObject})
+    vecs = Vector{Vector{Float64}}()
+    for state in states
+        vec = if isket(state) || isbra(state)
+            _state_to_bloch(state)
+        else
+            _dm_to_bloch(state)
+        end
+        push!(vecs, vec)
+    end
```
albertomercurio: Instead of using `push!`, you could do

```suggestion
vecs = map(states) do state
    if isket(state) || isbra(state)
        return _state_to_bloch(state)
    else
        return _dm_to_bloch(state)
    end
end
```

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +293,354 @@ raw"""
...
+    return add_vectors!(b, vecs)
```
albertomercurio: Same here, you are returning the same function with the same arguments. It can be misleading to the user.

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +293,354 @@ raw"""
...
+# Returns
+
+- A tuple `(fig, axis)` where `fig` is the figure object and `axis` is the axis object used for plotting.
+  These can be further manipulated or saved by the user.
+"""
+function QuantumToolbox.render(b::QuantumToolbox.Bloch; location = nothing)
+    fig, location = _getFigAndLocation(location)
+    if isnothing(b.fig) || b.fig !== fig
+        b.fig = fig
+        b.ax = Makie.Axis3(
```
albertomercurio: Even if the equivalent qutip class embeds also the fig and ax objects. I think that they are not needed in the `Bloch` struct. `render` would just generate fig and ax, without storing the in the `Bloch` object.

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +293,354 @@ raw"""
...
+            viewmode = :fit,
+        )
+        b.ax.azimuth[] = deg2rad(b.view_angles[1])
+        b.ax.elevation[] = deg2rad(b.view_angles[2])
+    else
+        if !(b.ax in contents(location))
+            location[] = b.ax
+        end
+    end
+    Makie.empty!(b.ax)
```
albertomercurio: Just a few things here.

The code is quite long. I would first write some explanation comments, and divide each operation as a separate function.

Moreover, it is not necessary to repeat `Makie.function`nif the function is imported with `using Makie: function`

src/visualization.jl
```diff
@@ -61,3 +71,340 @@ plot_fock_distribution(
...
+    vectors::Vector{Vector{Float64}}
+    lines::Vector{Tuple{Vector{Vector{Float64}},String,Dict{Any,Any}}}
+    arcs::Vector{Vector{Vector{Float64}}}
+    font_color::String
+    font_size::Int
+    frame_alpha::Float64
+    frame_color::String
+    frame_width::Int
+    point_default_color::Vector{String}
+    point_color::Vector{Any}
```
albertomercurio: I would avoid to use `Any`, as it makes type instabilities.

src/visualization.jl
```diff
@@ -61,3 +71,340 @@ plot_fock_distribution(
...
+    vector_width::Int
+    view_angles::Vector{Int}
+    xlabel::Vector{String}
+    xlpos::Vector{Float64}
+    ylabel::Vector{String}
+    ylpos::Vector{Float64}
+    zlabel::Vector{String}
+    zlpos::Vector{Float64}
+    fig::Any
+    ax::Any
```
albertomercurio: As I said before. fig and ax are not necessary to be stored in the `Bloch` structure. Once one calls `render`, they are generated and the user can operate on them.

src/visualization.jl
```diff
@@ -61,3 +71,340 @@ plot_fock_distribution(
...
+
+    function Bloch()
```
albertomercurio: Instead of defining this function, I would use `Base.`@kwdef`

https://docs.julialang.org/en/v1/base/base/#Base.@kwdef

ext/QuantumToolboxMakieExt.jl
albertomercurio: These functions, like `add_vectors!` or `add_lines!` seem to be quite short. Perhaps they can be directly embedded here.

src/visualization.jl
albertomercurio: Hi @Fe-r-oz, thank you. I have wrote some comments on the code. I have a few minor ones, which I will do after you apply these changes.

Anyhow, I see that now the arrows are too short. Isn't it?

test/runtests.jl
```diff
@@ -56,6 +56,7 @@ if (GROUP == "Makie_Ext")
     Pkg.develop(PackageSpec(path = dirname(@__DIR__)))
     Pkg.instantiate()
 
+    using CairoMakie
```
ytdHuang: This import need to be removed.

Cause we will test the error of missing import

The `CairoMakie` will be imported after that.

docs/make.jl
```diff
@@ -51,6 +51,7 @@ const PAGES = [
         ],
         "Manipulating States and Operators" => "users_guide/states_and_operators.md",
         "Tensor Products and Partial Traces" => "users_guide/tensor.md",
+        "Bloch Sphere Rendering" => "users_guide/plotting_the_bloch_sphere.md",
```
ytdHuang: We align most of the user guide structure with qutip, can you change the title as

```suggestion
        "Plotting on the Bloch Sphere" => "users_guide/plotting_the_bloch_sphere.md",
```

docs/src/users_guide/plotting_the_bloch_sphere.md
```diff
@@ -0,0 +1,102 @@
+# [Plotting on the Bloch Sphere](@id doc: plotting_the_bloch_sphere.md)
```
ytdHuang: We usually don't use space and underline to define `@id`s in docs.

Can you change it as

```suggestion
# [Plotting on the Bloch Sphere](@id doc:Plotting-on-the-Bloch-Sphere)
```

docs/src/users_guide/extensions/cairomakie.md
```diff
@@ -19,4 +19,5 @@ The supported plotting functions are listed as follows:
 | **Plotting Function** | **Description** |
 |:----------------------|:----------------|
 | [`plot_wigner`](@ref) | [Wigner quasipropability distribution](https://en.wikipedia.org/wiki/Wigner_quasiprobability_distribution) |
-| [`plot_fock_distribution`](@ref) | [Fock state](https://en.wikipedia.org/wiki/Fock_state) distribution |
\ No newline at end of file
+| [`plot_fock_distribution`](@ref) | [Fock state](https://en.wikipedia.org/wiki/Fock_state) distribution |
+| [`plot_bloch`](@ref)  | [Plotting on the Bloch Sphere](https://qutip.readthedocs.io/en/stable/guide/guide-bloch.html) |
```
ytdHuang: maybe we can just make the hyperlink here link to our own docs (as I changed the `@id` for the page below)

```suggestion
| [`plot_bloch`](@ref)  | [Plotting on the Bloch Sphere](@ref doc:Plotting-on-the-Bloch-Sphere) |
```

since we already mentioned the qutip Bloch sphere in that page

src/visualization.jl
```diff
@@ -1,4 +1,14 @@
...
+    plot_fock_distribution,
+    plot_bloch,
+    Bloch,
+    add_points!,
+    add_vectors!,
+    add_line!,
+    add_arc!,
+    clear!,
+    render,
+    add_states!
```
ytdHuang: small comment, better to separate the exports into groups to improve the readability of the code, something like:

```suggestion
export plot_wigner
export plot_fock_distribution
export plot_bloch,
    Bloch,
    render,
    add_points!,
    add_vectors!,
    add_line!,
    add_arc!,
    clear!,
    add_states!
```

src/visualization.jl
```diff
@@ -61,3 +71,301 @@ plot_fock_distribution(
 
 plot_fock_distribution(::Val{T}, ρ::QuantumObject{SType}; kwargs...) where {T,SType<:Union{Bra,Ket,Operator}} =
     throw(ArgumentError("The specified plotting library $T is not available. Try running `using $T` first."))
+
+@doc raw"""
+    Bloch()
+
+A structure representing a Bloch sphere visualization for quantum states."""
+@kwdef mutable struct Bloch
```
ytdHuang:
```suggestion
@doc raw"""
    Bloch()

A structure representing a Bloch sphere visualization for quantum states.
"""
@kwdef mutable struct Bloch
```

ytdHuang: (move the `"""` to a newline)

src/visualization.jl
```diff
@@ -61,3 +71,301 @@ plot_fock_distribution(
...
+    # Data storage
+    """Points to plot on the Bloch sphere (3D coordinates)"""
```
ytdHuang: I think these will not be showed in the `help>`

Can you gather all these descriptions for the fields into the docstring above?
like

```julia
@doc raw"""
    Bloch()

A structure representing a Bloch sphere visualization for quantum states.

# Fields

## Data storage
- `points::Vector{Matrix{Float64}}`: Points to plot on the Bloch sphere (3D coordinates)
- .
- .
- .

## Style properties
- `font_color::String`: Color of axis labels and text. Default to `"#2E3440"`
- .
- .
-.
"""

File: ext/QuantumToolboxMakieExt.jl [hunk 21]
```diff
@@ -291,4 +314,474 @@ raw"""
...
+Draw the three great circles `(XY, YZ, XZ planes)` on the Bloch sphere.
+
+# Arguments
+- `ax`: Makie Axis3 object for drawing
+
+Adds faint circular guidelines representing the three principal planes.
+"""
+function _draw_reference_circles(ax)
+    wire_color = RGBAf(0.5, 0.5, 0.5, 0.4)
+    φ = range(0, 2π, length = 100)
```
albertomercurio, line 586: By dividing into functions helped me a lot on the readability. What happens if we reduce the length of φ for example to 10? Do we reduce the number of circles? If so, then I prefer to have less circles, as said a few comments ago. Could you plot an example here?

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +314,474 @@ raw"""
...
+
+Creates visible axis lines extending slightly beyond the unit sphere.
+"""
+function _draw_axes(ax)
+    axis_color = RGBAf(0.3, 0.3, 0.3, 0.8)
+    axis_width = 0.8
+    axes = [
+        ([Point3f(0, -1.01, 0), Point3f(0, 1.01, 0)], "y"),  # Y-axis
+        ([Point3f(-1.01, 0, 0), Point3f(1.01, 0, 0)], "x"),  # X-axis
+        ([Point3f(0, 0, -1.01), Point3f(0, 0, 1.01)], "z"),  # Z-axis
```
albertomercurio: Why 1.01 and not simply 1.0? You can send some examples here in case.

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +314,474 @@ raw"""
...
+function QuantumToolbox.render(b::Bloch; location = nothing)
+    fig, ax = _setup_bloch_plot(b, location)
+    _draw_bloch_sphere(ax)
+    _draw_reference_circles(ax)
+    _draw_axes(ax)
+    _plot_points(b, ax)
+    _plot_lines(b, ax)
+    _plot_arcs(b, ax)
+    _plot_vectors(b, ax)
+    _add_labels(ax)
```
albertomercurio: Fantastic, thanks. Could you just add the `!` symbol to tell this converts `ax` inplace?

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +314,474 @@ raw"""
...
+function _add_labels(ax)
+    label_color = RGBf(0.2, 0.2, 0.2)
+    label_size = 16
+    label_font = "TeX Gyre Heros"
+
+    labels = [
+        (Point3f(1.04, 0, 0), L"\textbf{y}"),
+        (Point3f(0, -1.10, 0), L"\textbf{x}"),
+        (Point3f(0, 0, -1.10), L"\mathbf{|1\rangle}"),
+        (Point3f(0, 0, 1.08), L"\mathbf{|0\rangle}"),
```
albertomercurio: The labels are a bit too close to the figure. Could you move them out a bit? Also, why `y` and `x` are `textbf` while in the other we have `mathbf`?

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +314,474 @@ raw"""
...
+    ]
+    for (pos, text) in labels
+        text!(ax, text; position = pos, color = label_color, fontsize = label_size, font = label_font)
+    end
+end
+
+function QuantumToolbox.plot_bloch(::Val{:Makie}, state::QuantumObject{<:Union{Ket,Bra}}; kwargs...)
+    state = isbra(state) ? dag(state) : state
```
albertomercurio: Using Julia built-in function

```suggestion
    state = isbra(state) ? state' : state
```

src/visualization.jl
```diff
@@ -61,3 +71,301 @@ plot_fock_distribution(
...
+    points::Vector{Matrix{Float64}} = Vector{Matrix{Float64}}()
+    """Vectors to plot on the Bloch sphere"""
+    vectors::Vector{Vector{Float64}} = Vector{Vector{Float64}}()
+    """Lines to draw on the sphere (points, style, properties)"""
+    lines::Vector{Tuple{Vector{Vector{Float64}},String,Dict{Symbol,Any}}} =
```
albertomercurio: Why do we have `Dict{Symbol, Any}`? Can't we infer the type?

ext/QuantumToolboxMakieExt.jl
```diff
@@ -291,4 +314,474 @@ raw"""
...
+
+Draw the translucent sphere representing the Bloch sphere surface.
+
+# Arguments
+- `ax`: Makie Axis3 object where the sphere will be drawn
+
+Creates a semi-transparent spherical surface at the origin with radius 1.
+"""
+function _draw_bloch_sphere(ax)
+    sphere_color = RGBAf(1.0, 0.86, 0.86, 0.2)
```
albertomercurio: In the `Bloch` structure we have defined the `sphere_alpha` field. But it seems to not been used here. Am I wrong?

albertomercurio: In general, let's check that we use all the fields of `Bloch`.

test/ext-test/cpu/makie/makie_ext.jl
```diff
@@ -60,4 +60,113 @@
     fig = Figure()
     pos = fig[2, 3]
     fig1, ax = @test_logs (:warn,) plot_fock_distribution(ψ * 2; library = Val(:Makie), location = pos)
+
```
albertomercurio, line 64: They are quite a lot, which is nice, but can you embed them inside a `@testset`?

ext/QuantumToolboxMakieExt.jl
Fe-r-oz: Thank you for your patience and for raising this important point again. Sorry for the delay in addressing this suggestion properly. Regarding the implementation, it appears that `Makie.GeometryBasics.Sphere` (center + radius) wasn’t suitable for our needs, so I’ve built a custom sphere via `surface!` instead. Please let me know your feedback!!

The previous approach:
when φ = 10
display

when φ = 100
display

The new approach:

 > Edit: Temporarily using this color scheme as the default.

display

display

display

display

display

Since we are using `surface!`,  lighter tones combined with transparency may reveal the underlying mesh grid of the `surface!`.

display

display

albertomercurio: How did you generate the sphere with the other approach? Could the `mesh!` function together with `Sphere` work?

ext/QuantumToolboxMakieExt.jl
albertomercurio: In general, they are also not aligned to the axes. They should be the continuation of the corresponding axis

ext/QuantumToolboxMakieExt.jl
Fe-r-oz: I generated the sphere using a custom parametric mesh via the `surface!` function, rather than relying on the built-in `Sphere` type with `mesh!`. This approach gives us precise control over the number and placement of longitude and latitude lines. The sphere is explicitly constructed from parametric equations in spherical coordinates.

While Makie’s `Sphere` type (an alias for [`HyperSphere{3, T}`](https://github.com/JuliaGeometry/GeometryBasics.jl/blob/caa6c89f9dabe2512fb1ab8095d75aff9f9e4ea3/src/primitives/spheres.jl#L17)) is convenient for basic 3D geometry, it has certain design constraints. For example, mesh generation is fixed internally using [`coordinates(s::Sphere, nvertices=24)`](https://github.com/JuliaGeometry/GeometryBasics.jl/blob/caa6c89f9dabe2512fb1ab8095d75aff9f9e4ea3/src/primitives/spheres.jl#L43), which samples spherical angles using a hardcoded resolution. Unfortunately, it appears that `nvertices` isn't configurable through the `mesh!` interface. I also explored modifying the `Sphere`'s `mesh `parameters directly, but it seems this flexibility isn’t currently exposed through the high-level API.

albertomercurio: But can't you play with transparency and alpha on the mesh? What I don't like of this surface method is that I can see the underlying mesh grid.

Fe-r-oz: We can indeed tweak the alpha and transparency settings to reduce or eliminate the appearance of the mesh grid when using `surface!`. The last two examples I shared earlier were mostly exploratory, just to get a feel for different visual effects while playing with color and opacity settings. I also don't like the underlying grid look  which was likely caused by the alpha and transparency settings.

For example, tweaking alpha and transparency settings, we get:

display
display

> Edit: increasing the resolution, the mesh grid disappears. Some examples; 

display
display


We can refine this method further; I will explore additional tweaks and see how to achieve a cleaner result with `surface!` Please let me know your feedback! Thank you!

albertomercurio: I think the first image is really nice. How did you get that?

Fe-r-oz: Thank you! There are two ways of generating the first image. We can use both `sphere + mesh` or `surface!`. I generated the above images via `surface!`. But I have been following through your comment and as you mentioned, we can do the same with `sphere+mesh` (figure attached below). The goal is to disappear the mesh grid, and draw the few latitude and longitude lines ourselves.


`Sphere+mesh` approach:
display

> Edit: more polished version

display

albertomercurio: Ok, I like it. This could be the definitive one.

albertomercurio:
I was also thinking, could you also test it with GLMakie instead of CairoMakie? This should be done without any change, as we support Makie in general. I'm curious to see if the rendering quality is the same and what happens in the interactive windows, since now we can move interactively

Fe-r-oz:
Bloch sphere using `GLMakie` using the current pushed version 

Screenshot_select-area_20250603133522

It appears ` :rect, :diamond, :utriangle` shapes for scatter of points don't appear to render well using `GLMakie`
Screenshot_select-area_20250603133436

.typos.toml
```diff
@@ -1,3 +1,4 @@
 [default.extend-words]
 ket = "ket"
 sme = "sme"
+Heros = "Heros"
```
ytdHuang: can we change the font family `"TeX Gyre Heros"` ?

so that we don't need to add this?

For the LaTeX labels, I have another way to do it, as shown in the next comment

src/visualization.jl
```diff
@@ -61,3 +71,345 @@ plot_fock_distribution(
...
+
+- `view_angles::Tuple{Int,Int}}`: Azimuthal and elevation viewing angles in degrees (default: (-60, 30))
+
+## Label properties
+- `xlabel::Vector{String}}`: Labels for x-axis (default: ["x", ""])
+- `xlpos::Vector{Float64}}`: Positions of x-axis labels (default: [1.2, -1.2])
+- `ylabel::Vector{String}}`: Labels for y-axis (default: ["y", ""])
+- `ylpos::Vector{Float64}}`: Positions of y-axis labels (default: [1.2, -1.2])
+- `zlabel::Vector{String}}`: Labels for z-axis (default: ["|0⟩", "|1⟩"])
+- `zlpos::Vector{Float64}}`: Positions of z-axis labels (default: [1.2, -1.2])
```
ytdHuang: for the `label`s, I think we can make the type as `Vector{AbstractString}`

so that the default values can use `LaTexString` :
- `[L"x", ""]`
- `[L"y", ""]`
- `[L"|0\rangle", L"|1\rangle"]`

src/visualization.jl
```diff
@@ -61,3 +71,345 @@ plot_fock_distribution(
...
+    vector_color::Vector{String} = ["green", "orange", "blue", "red"]
+    vector_width::Float64 = 0.025
+    vector_arrowsize::NTuple{3,Real} = (0.07, 0.08, 0.08)
+    view_angles::Tuple{Int,Int} = (-60, 30)
+    xlabel::Vector{String} = ["x", ""]
+    xlpos::Vector{Float64} = [1.3, -1.3]
+    ylabel::Vector{String} = ["y", ""]
+    ylpos::Vector{Float64} = [1.3, -1.3]
+    zlabel::Vector{String} = ["|0⟩", "|1⟩"]
+    zlpos::Vector{Float64} = [1.2, -1.3]
```
ytdHuang: @albertomercurio 

Do you think we should set most of the default values here to be the same as qutip (https://qutip.readthedocs.io/en/latest/guide/guide-bloch.html) ?

the settings that effects the color and style of the Bloch sphere, such as `frame_color`, `frame_alpha`, `sphere_color`, `xlpos`, `ylpos`, `zlpos`, etc.

ytdHuang: my point is to make the default style of the Bloch sphere looks as similar as the one in qutip

Fe-r-oz: Thank you! I used the qutip hex color and labels positions, and shifted to Ariel font family: 

display

It seems that previous font was `Makie`'s default one: 

```
using Makie
Makie.theme(:fonts)
Attributes with 4 entries:
  bold => TeX Gyre Heros Makie Bold
```

> Edit: Makie does not support Arial, so I have removed this font name attribute as it is not present in qutip documentation well.

src/visualization.jl
Fe-r-oz: Thank you!!

src/visualization.jl
```diff
@@ -61,3 +71,343 @@ plot_fock_distribution(
...
+
+- `view_angles::Tuple{Int,Int}}`: Azimuthal and elevation viewing angles in degrees (default: (-60, 30))
+
+## Label properties
+- `xlabel::Vector{String}}`: Labels for x-axis (default: ["x", ""])
+- `xlpos::Vector{Float64}}`: Positions of x-axis labels (default: [1.0, -1.0])
+- `ylabel::Vector{String}}`: Labels for y-axis (default: ["y", ""])
+- `ylpos::Vector{Float64}}`: Positions of y-axis labels (default: [1.0, -1.0])
+- `zlabel::Vector{String}}`: Labels for z-axis (default: ["|0⟩", "|1⟩"])
+- `zlpos::Vector{Float64}}`: Positions of z-axis labels (default: [1.0, -1.0])
```
ytdHuang:
```suggestion
- `xlabel::Vector{AbstractString}}`: Labels for x-axis (default: [L"x", ""])
- `xlpos::Vector{Float64}}`: Positions of x-axis labels (default: [1.0, -1.0])
- `ylabel::Vector{AbstractString}}`: Labels for y-axis (default: [L"y", ""])
- `ylpos::Vector{Float64}}`: Positions of y-axis labels (default: [1.0, -1.0])
- `zlabel::Vector{AbstractString}}`: Labels for z-axis (default: [L"|0\rangle", L"|1\rangle"])
- `zlpos::Vector{Float64}}`: Positions of z-axis labels (default: [1.0, -1.0])
```

src/visualization.jl
```diff
@@ -61,3 +71,343 @@ plot_fock_distribution(
...
+    vector_color::Vector{String} = ["green", "orange", "blue", "red"]
+    vector_width::Float64 = 0.025
+    vector_arrowsize::NTuple{3,Real} = (0.07, 0.08, 0.08)
+    view_angles::Tuple{Int,Int} = (-60, 30)
+    xlabel::Vector{AbstractString} = ["x", ""]
+    xlpos::Vector{Float64} = [1.0, -1.0]
+    ylabel::Vector{AbstractString} = ["y", ""]
+    ylpos::Vector{Float64} = [1.0, -1.0]
+    zlabel::Vector{AbstractString} = ["|0⟩", "|1⟩"]
+    zlpos::Vector{Float64} = [1.0, -1.0]
```
ytdHuang:
```suggestion
    xlabel::Vector{AbstractString} = [L"x", ""]
    xlpos::Vector{Float64} = [1.0, -1.0]
    ylabel::Vector{AbstractString} = [L"y", ""]
    ylpos::Vector{Float64} = [1.0, -1.0]
    zlabel::Vector{AbstractString} = [L"|0\rangle", L"|1\rangle"]
    zlpos::Vector{Float64} = [1.0, -1.0]
```

Fe-r-oz:
I did `make docs/format/changelog/test` and the documentation build passed locally. Hopefully, the CI runs will pass here as well! Thank you for reviewing this PR - I appreciate your time and feedback!!

src/visualization.jl
ytdHuang: ok

ytdHuang:
Ok, thank you for your hard work.
I will trigger the CI pipeline.

src/visualization.jl
Fe-r-oz: It appears that LaTeXStrings `@L_str` are not used outside of `Makie` Ext so this causes error as we use `L` in `visualization.jl`

src/visualization.jl
albertomercurio: It is really nice like that

ytdHuang (approved):

PR #476, opened 2025-05-28
Lanczos-based `spectrum` method

### LINKED ISSUE #463: Implement Lanczos-based `spectrum` method
For a given steady-state correlation function $\left\langle \hat{A}(\tau) \hat{B}(0) \right\rangle$, [`spectrum`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.spectrum) calculates the quantity
```math
S(\omega) = \int_{-\infty}^\infty \left\langle \hat{A}(\tau) \hat{B}(0) \right\rangle e^{-i \omega \tau} d \tau
```
— see also the documentation on [Emission spectrum](https://qutip.org/QuantumToolbox.jl/stable/users_guide/two_time_corr_func#Emission-spectrum).

Compared to [`spectrum_correlation_fft`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.spectrum_correlation_fft), [`spectrum`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.spectrum)  works directly in the frequency domain by using a `SpectrumSolver`. In fact, in the vectorized representation, $S(\omega)$ can be expressed as
```math
S(\omega) = 2 \mathrm{Im}
\bigg\langle I \bigg\vert \hat{A}
\frac{1}{\omega + i\mathcal{L}}
\hat{B} \bigg\vert \rho_{\mathrm{ss}} \bigg\rangle
```

Currently, there two such solvers:
- [`ExponentialSeries`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.ExponentialSeries), which employes a full spectral decomposition;
- [`PseudoInverse`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.PseudoInverse), which computes the propagator $(\omega + i\mathcal{L})^{-1}$ as the solution of a sparse linear system at each frequency.

[`PseudoInverse`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.PseudoInverse) is currently the fastest option for large systems, but it requires (pseudo-)inverting $\omega + i\mathcal{L}$ at each frequency value.

Another possible avenue for the computation of the spectrum would be to employ the well-known Lanczos method, which is currently a gold standard in condensed matter calculations. The method is extensively discussed in <a name="backrefKoch2011"></a> [[1]](#refKoch2011) for systems governed by a Hamiltonian $H$, in which response functions are calculated on top of the ground state $\vert \psi_{0} \rangle$:
```math
\begin{equation}
    S(\omega)
    = 2  \mathrm{Im} \Braket{\psi_0 | \hat{A}\frac{1}{\omega-\hat{H}}\hat{B} | \psi_0}
\end{equation}
```

In a nutshell, for $\hat{A} = \hat{B}^{\dagger}$, $H$ is iteratively expressed in a Krylov basis built on top of $\vert v_{1} \rangle = \hat{B} \vert \psi_{0} \rangle$ by successive applications of $H$. In this basis the Hamiltonian becomes tridiagonal:
```math
\begin{equation}
    \hat{H}_{\mathcal{K}(\ket{v_1})} =
    \begin{pmatrix}
        a_1    & b_2 & 0   & 0   & \cdots \\
        b_2    & a_2 & b_3 & 0   &        \\
        0      & b_3 & a_3 & b_4 &        \\
        0      & 0   & b_4 & a_4 &        \\
        \vdots &     &     &     & \ddots
    \end{pmatrix}\,.
\end{equation}
```
Therefore, the desidered expectation value of the propagator $(\omega - H)^{-1}$ can be read off its top-left matrix element $\left( \omega - \hat{H} \right)^{-1}_{11}$.

Thanks to the tridiagonal structure of the propagator, the top-left matrix element can be recursively expressed as a continuous fraction for each Krylov iteration by using a suitable expression for the inverse of a $2 \times 2$ block matrix:
```math
\begin{equation}
    \left( \omega - \hat{H} \right)^{-1}_{11} 
    = \cfrac{1}{\omega-a_1-\cfrac{b_2^2}{\omega-a_2-\cfrac{b_3^2}{\omega-a_3-\cdots}}} \,.
\end{equation}
```

The algorithm stops when the continuous fraction reaches the desired precision or the maximum number of iterations. Early stops are possible in case of loss of orthogonality, a known drawback of the Lanczos algorithm; however, they are not as catastrophic as for the case of pure diagonalization, they can be easily detected as $b_n$ becomes $0$, and they just result in a termination of the continued fraction.

Cases in which $\hat{A} \neq \hat{B}^{\dagger}$ can be handled by starting the Krylov basis from $(\hat{B} \pm \hat{A}^{\dagger}) \vert \psi_{0} \rangle$ and expanding the sums in the expectation value.

In our case, the goal would be to implement a `SpectrumSolver` for [`spectrum`](https://qutip.org/QuantumToolbox.jl/stable/resources/api#QuantumToolbox.spectrum) called **`Lanczos`**, which implements a non-symmetric variant of the algorithm due to the fact that $\mathcal{L}$, unlike $\mathcal{H}$, is non-Hermitian. It should support a usage like:
```julia
spectrum(H, ωlist, c_ops, A, B; solver = Lanczos())
```
and provide, at a minimum, a `tol` parameter and a `maxiter` parameter, respectively controlling the desired precision and the maximum number of Krylov iterations.

---

1. <a name="refKoch2011"></a> Koch, E. (2011). [The Lanczos Method](https://www.cond-mat.de/events/correl11/manuscripts/koch.pdf). In E. Pavarini, E. Koch, D. Vollhardt, & A. Lichtenstein (Eds.), _The LDA+DMFT approach to strongly correlated materials_ (Vol. 1, pp. 8.1-8.30). Forschungszentrum Jülich. [↩](#backrefKoch2011)

matteosecli:
## Checklist

## Description
Lanczos-based `spectrum` method.

## Related issues or PRs
Resolves #463.

## Additional context
This PR implements the non-symmetric variant of the Hermitian Lanczos method described in [Koch2011](https://www.cond-mat.de/events/correl11/manuscripts/koch.pdf).

The pseudo-code for the non-symmetric Lanczos algorithm can be found in Algorithm 6.6 of [Saad2011](https://www-users.cse.umn.edu/~saad/eig_book_2ndEd.pdf).

At each step, the running estimate for the result is updated via a [Wallis-Euler recursion](https://en.wikipedia.org/wiki/Continued_fraction) to avoid recomputing the whole continued fraction each time, and the convergence is evaluated accordingly.

### DIFF
--- a/src/spectrum.jl
+++ b/src/spectrum.jl
@@ -33,10 +33,10 @@ A solver which solves [`spectrum`](@ref) by using a non-symmetric Lanczos varian
 The nonsymmetric Lanczos algorithm is adapted from Algorithm 6.6 in [Saad2011](https://www-users.cse.umn.edu/~saad/eig_book_2ndEd.pdf).
 The running estimate is updated via a [Wallis-Euler recursion](https://en.wikipedia.org/wiki/Continued_fraction).
 """
-Base.@kwdef struct Lanczos{T<:Real,IT<:Int,SS<:Union{Nothing,<:SteadyStateSolver}} <: SpectrumSolver
+Base.@kwdef struct Lanczos{T<:Real,SS<:Union{Nothing,<:SteadyStateSolver}} <: SpectrumSolver
     tol::T = 1e-8
-    maxiter::IT = 5000
-    verbose::IT = 0
+    maxiter::Int = 5000
+    verbose::Int = 0
     steadystate_solver::SS = nothing
 end
 
@@ -223,7 +223,7 @@ function _spectrum(
     w₊₁ = vT(zeros(cT, D^2))
 
     # Frequency of renormalization
-    renormFrequency::typeof(solver.maxiter) = 1
+    renormFrequency = 1
 
     # Loop over the Krylov subspace(s)
     for k in 1:solver.maxiter
--- a/src/spectrum.jl
+++ b/src/spectrum.jl
@@ -33,10 +33,11 @@ A solver which solves [`spectrum`](@ref) by using a non-symmetric Lanczos varian
 The nonsymmetric Lanczos algorithm is adapted from Algorithm 6.6 in [Saad2011](https://www-users.cse.umn.edu/~saad/eig_book_2ndEd.pdf).
 The running estimate is updated via a [Wallis-Euler recursion](https://en.wikipedia.org/wiki/Continued_fraction).
 """
-Base.@kwdef struct Lanczos{T<:Real,IT<:Int} <: SpectrumSolver
+Base.@kwdef mutable struct Lanczos{T<:Real,IT<:Int} <: SpectrumSolver
     tol::T = 1e-8
     maxiter::IT = 5000
     verbose::IT = 0
+    steadystate_solver::Union{Nothing,SteadyStateSolver} = nothing
 end
 
 @doc raw"""
@@ -172,7 +173,9 @@ function _spectrum(
     cT = _CType(L)
 
     # Calculate |v₁> = B|ρss>
-    ρss = mat2vec(steadystate(L))
+    ρss =
+        solver.steadystate_solver === nothing ? mat2vec(steadystate(L)) :
+        mat2vec(steadystate(L; solver = solver.steadystate_solver))
     vₖ = (spre(B) * ρss).data
 
     # Define (possibly GPU) vector type
--- a/src/spectrum.jl
+++ b/src/spectrum.jl
@@ -174,7 +174,7 @@ function _spectrum(
 
     # Calculate |v₁> = B|ρss>
     ρss =
-        solver.steadystate_solver === nothing ? mat2vec(steadystate(L)) :
+        isnothing(solver.steadystate_solver) ? mat2vec(steadystate(L)) :
         mat2vec(steadystate(L; solver = solver.steadystate_solver))
     vₖ = (spre(B) * ρss).data
 
@@ -219,8 +219,8 @@ function _spectrum(
     # Previous and next left/right Krylov vectors
     v₋₁ = vT(zeros(cT, D^2))
     v₊₁ = vT(zeros(cT, D^2))
-    w₋₁ = vT(zeros(cT, D^2))'
-    w₊₁ = vT(zeros(cT, D^2))'
+    w₋₁ = transpose(vT(zeros(cT, D^2)))
+    w₊₁ = transpose(vT(zeros(cT, D^2)))
 
     # Frequency of renormalization
     renormFrequency::typeof(solver.maxiter) = 1
@@ -228,7 +228,7 @@ function _spectrum(
     # Loop over the Krylov subspace(s)
     for k in 1:solver.maxiter
         # k-th diagonal element
-        w₊₁ = wₖ * L.data
+        mul!(w₊₁.parent, transpose(L.data), wₖ.parent) # Equivalent to: w₊₁ = wₖ * L.data
         αₖ = w₊₁ * vₖ
 
         # Update A(k), B(k) and continuous fraction; normalization avoids overflow
@@ -239,7 +239,7 @@ function _spectrum(
 
         # Renormalize Euler sequences to avoid overflow
         if k % renormFrequency == 0
-            maxNorm .= max.(abs.(Aₖ), abs.(Bₖ))
+            map!((x, y) -> max(abs(x), abs(y)), maxNorm, Aₖ, Bₖ)
             Aₖ ./= maxNorm
             Bₖ ./= maxNorm
             A₋₁ ./= maxNorm
@@ -248,8 +248,8 @@ function _spectrum(
 
         # Check for convergence
         maxResidue =
-            maximum(abs.(lanczosFactor .- lanczosFactor₋₁)) /
-            max(maximum(abs.(lanczosFactor)), maximum(abs.(lanczosFactor₋₁)))
+            mapreduce((x, y) -> abs(x - y), max, lanczosFactor, lanczosFactor₋₁) /
+            max(maximum(abs, lanczosFactor), maximum(abs, lanczosFactor₋₁))
         if maxResidue <= solver.tol
             if solver.verbose > 1
                 println("spectrum(): solver::Lanczos converged after $(k) iterations")
@@ -258,14 +258,13 @@ function _spectrum(
         end
 
         # (k+1)-th left/right vectors, orthogonal to previous ones
-        # Consider using explicit BLAS calls
-        v₊₁ = L.data * vₖ
+        mul!(v₊₁,L.data,vₖ) # Equivalent to: v₊₁ = L.data * vₖ
         v₊₁ .= v₊₁ .- αₖ .* vₖ .- βₖ .* v₋₁
         w₊₁ .= w₊₁ .- αₖ .* wₖ .- δₖ .* w₋₁
-        v₋₁ .= vₖ
-        w₋₁ .= wₖ
-        vₖ .= v₊₁
-        wₖ .= w₊₁
+        v₋₁, vₖ = vₖ, v₋₁
+        vₖ, v₊₁ = v₊₁, vₖ
+        w₋₁, wₖ = wₖ, w₋₁
+        wₖ, w₊₁ = w₊₁, wₖ
 
         # k-th off-diagonal elements
         buf = wₖ * vₖ
@@ -277,10 +276,10 @@ function _spectrum(
         wₖ ./= βₖ
 
         # Update everything for the next cycle
-        A₋₂ .= A₋₁
-        A₋₁ .= Aₖ
-        B₋₂ .= B₋₁
-        B₋₁ .= Bₖ
+        A₋₂, A₋₁ = A₋₁, A₋₂
+        A₋₁, Aₖ = Aₖ, A₋₁
+        B₋₂, B₋₁ = B₋₁, B₋₂
+        B₋₁, Bₖ = Bₖ, B₋₁
     end
 
     if solver.verbose > 0 && maxResidue > solver.tol
--- a/src/spectrum.jl
+++ b/src/spectrum.jl
@@ -193,24 +193,24 @@ function _spectrum(
     Length = length(ωList)
 
     # Current and previous estimates of the spectrum
-    lanczosFactor   = vT(zeros(cT, Length))
+    lanczosFactor = vT(zeros(cT, Length))
     lanczosFactor₋₁ = vT(zeros(cT, Length))
 
     # Tridiagonal matrix elements
-    αₖ = cT( 0)
+    αₖ = cT(0)
     βₖ = cT(-1)
     δₖ = cT(+1)
 
     # Current and up to second-to-last A and B Euler sequences
-    A₋₂ = vT( ones(cT, Length))
+    A₋₂ = vT(ones(cT, Length))
     A₋₁ = vT(zeros(cT, Length))
-    Aₖ  = vT(zeros(cT, Length))
+    Aₖ = vT(zeros(cT, Length))
     B₋₂ = vT(zeros(cT, Length))
-    B₋₁ = vT( ones(cT, Length))
-    Bₖ  = vT(zeros(cT, Length))
+    B₋₁ = vT(ones(cT, Length))
+    Bₖ = vT(zeros(cT, Length))
 
     # Maximum norm and residue
-    maxNorm    = vT(zeros(cT, length(ωList)))
+    maxNorm = vT(zeros(cT, length(ωList)))
     maxResidue = fT(0.0)
 
     # Previous and next left/right Krylov vectors
@@ -227,7 +227,7 @@ function _spectrum(
         # k-th diagonal element
         w₊₁ = wₖ * L.data
         αₖ = w₊₁ * vₖ
-        
+
         # Update A(k), B(k) and continuous fraction; normalization avoids overflow
         Aₖ .= (-1im .* ωList .+ αₖ) .* A₋₁ .- (βₖ * δₖ) .* A₋₂
         Bₖ .= (-1im .* ωList .+ αₖ) .* B₋₁ .- (βₖ * δₖ) .* B₋₂
@@ -236,7 +236,7 @@ function _spectrum(
 
         # Renormalize Euler sequences to avoid overflow
         if k % renormFrequency == 0
-            maxNorm .= max.(abs.(Aₖ), abs.(Bₖ))  # Note: the MATLAB and C++ codes return the actual complex number
+            maxNorm .= max.(abs.(Aₖ), abs.(Bₖ))
             Aₖ ./= maxNorm
             Bₖ ./= maxNorm
             A₋₁ ./= maxNorm
@@ -244,8 +244,9 @@ function _spectrum(
         end
 
         # Check for convergence
-        maxResidue = maximum(abs.(lanczosFactor .- lanczosFactor₋₁)) / 
-                     max(maximum(abs.(lanczosFactor)), maximum(abs.(lanczosFactor₋₁)))
+        maxResidue =
+            maximum(abs.(lanczosFactor .- lanczosFactor₋₁)) /
+            max(maximum(abs.(lanczosFactor)), maximum(abs.(lanczosFactor₋₁)))
         if maxResidue <= solver.tol
             if solver.verbose > 1
                 println("spectrum(): solver::Lanczos converged after $(k) iterations")
@@ -260,13 +261,13 @@ function _spectrum(
         w₊₁ .= w₊₁ .- αₖ .* wₖ .- δₖ .* w₋₁
         v₋₁ .= vₖ
         w₋₁ .= wₖ
-        vₖ  .= v₊₁
-        wₖ  .= w₊₁
+        vₖ .= v₊₁
+        wₖ .= w₊₁
 
         # k-th off-diagonal elements
         buf = wₖ * vₖ
-        δₖ  = sqrt(abs(buf))
-        βₖ  = buf / δₖ
+        δₖ = sqrt(abs(buf))
+        βₖ = buf / δₖ
 
         # Normalize (k+1)-th left/right vectors
         vₖ ./= δₖ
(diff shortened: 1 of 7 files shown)


matteosecli:
I've been running some benchmarks (on CPU) in the meantime.

I'm just using the example for the vacuum Rabi splitting, with the following script that writes out the runtime (minus compile time) out in the legend:

<details>

```julia
using QuantumToolbox
using CairoMakie

# %%
N  = 4           # number of cavity fock states
ωc = 1.0 * 2 * π  # cavity frequency
ωa = 1.0 * 2 * π  # atom frequency
g  = 0.1 * 2 * π  # coupling strength
κ  = 0.75         # cavity dissipation rate
γ  = 0.25         # atom dissipation rate

# Jaynes-Cummings Hamiltonian
a  = tensor(destroy(N), qeye(2))
sm = tensor(qeye(N), destroy(2))
H = ωc * a' * a + ωa * sm' * sm + g * (a' * sm + a * sm')

# collapse operators
n_th = 0.25
c_ops = [
    sqrt(κ * (1 + n_th)) * a,
    sqrt(κ *      n_th)  * a',
    sqrt(γ)              * sm,
];

# calculate the correlation function using mesolve, and then FFT to obtain the spectrum.
# Here we need to make sure to evaluate the correlation function for a sufficient long time and
# sufficiently high sampling rate so that FFT captures all the features in the resulting spectrum.
tlist = LinRange(0, 100, 5000)
corr = correlation_2op_1t(H, nothing, tlist, c_ops, a', a; progress_bar = Val(false))
ωlist1, spec1 = spectrum_correlation_fft(tlist, corr)

# calculate the power spectrum using spectrum
# using Exponential Series (default) method
ωlist2 = LinRange(0.25, 1.75, 200) * 2 * π
(spec2, rt2, _, _, _, _, ct2, _) = @timed spectrum(H, ωlist2, c_ops, a', a; solver = ExponentialSeries())

# calculate the power spectrum using spectrum
# using Pseudo-Inverse method
(spec3, rt3, _, _, _, _, ct3, _) = @timed spectrum(H, ωlist2, c_ops, a', a; solver = PseudoInverse())

# calculate the power spectrum using spectrum
# using Lanczos method
(spec4, rt4, _, _, _, _, ct4, _) = @timed spectrum(H, ωlist2, c_ops, a', a; solver = Lanczos())

# plot by CairoMakie.jl
fig = Figure(size = (600, 420))
ax = Axis(fig[1, 1], title = "Vacuum Rabi splitting", xlabel = "Frequency", ylabel = "Emission power spectrum")
lines!(ax, ωlist1 / (2 * π), spec1, label = "mesolve + FFT", linestyle = :solid)
lines!(ax, ωlist2 / (2 * π), spec2, label = string(rt2-ct2)*": "*"Exponential Series", linestyle = :dash)
lines!(ax, ωlist2 / (2 * π), spec3, label = string(rt3-ct3)*": "*"Pseudo-Inverse", linestyle = :dashdot)
lines!(ax, ωlist2 / (2 * π), spec4, label = string(rt4-ct4)*": "*"Lanczos", linestyle = :dot)

xlims!(ax, ωlist2[1] / (2 * π), ωlist2[end] / (2 * π))
axislegend(ax, position = :rb)

fig

# save the figure
#save("vacuumRabiSplitting.png", fig)
#save("vacuumRabiSplitting.pdf", fig)
```

</details>

I've started by increasing the cutoff to (a useless) `N=16`, just to be able to appreciate the difference:

vrs_N16_w200

`PseudoInverse` is 10 times faster than `ExponentialSeries`, and in turn `Lanczos` is `42` (no pun intended) times faster than `PseudoInverse`.

But it shines even more if you try to increase your frequency resolution, for example by asking for `5000` points instead of `200` (again, useless in this case, but it's just for demonstration purposes):

vrs_N16_w5000

Now `PseudoInverse` actually takes twice as much as `ExponentialSeries`, becoming 22 times slower, while `Lanczos` slows down only by a factor 5, thus being almost 200 times faster than `PseudoInverse` and more than 90 times faster than `ExponentialSeries`.

If you then try to increase the cutoff too, to e.g. an even more useless `N=32`, `Lanczos` becomes more than 3500 times faster than `ExponentialSeries` and 730 times faster than `PseudoInverse`:

vrs_N32_w5000

In summary, at least according to these initial tests, while `ExponentialSeries` struggles with higher cutoffs and `PseudoInverse` struggles with higher frequency resolution (because it has to compute a pseudo-inverse for each frequency point), `Lanczos` seems to offer a competitive advantage in both scenarios.

albertomercurio:
Nice! I will give a look later.

src/spectrum.jl
```diff
@@ -144,6 +160,143 @@ function _spectrum(
...
+
+        # Update everything for the next cycle
+        A₋₂ .= A₋₁
+        A₋₁ .= Aₖ
+        B₋₂ .= B₋₁
+        B₋₁ .= Bₖ
+    end
+
+    if solver.verbose > 0
+        if maxResidue > solver.tol
```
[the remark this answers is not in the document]
matteosecli: I wanted to save repeated evaluations of `maxResidue > solver.tol`, and evaluate it only if the user asks information about convergence, but I guess it's so inexpensive I can as well do it at every step as you suggested, I'll change that :)

src/spectrum.jl
```diff
@@ -144,6 +160,143 @@ function _spectrum(
     return spec
 end
 
+function _spectrum(
+    L::QuantumObject{SuperOperator},
+    ωlist::AbstractVector,
+    A::QuantumObject{Operator},
+    B::QuantumObject{Operator},
+    solver::Lanczos;
+    kwargs...,
```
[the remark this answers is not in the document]
ytdHuang: I was just thinking it is more intuitive that `verbose` is a `kwargs`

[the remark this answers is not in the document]
matteosecli: Sure, I'll remove that. `verbose` is Lanczos-specific; since it's an iterative method, I thought it could be useful to have convergence information.

Specifically, if `verbose=0`, the algorithm will return whatever it calculated without warning the user about non-convergence. If `verbose=1`, it will instead warn the user, print out the current residue (that should instead be lower than the user-specified tolerance), and suggest to increase the number of iterations (which in many cases fixes the issue). If `verbose>1`, it will also print information about the actual number of iterations carried out when the algorithm converges. Higher verbosity numbers could be implemented later if necessary. For example, the previous behavior could be assigned `verbose=2`, while `verbose=3` could print the residue at each iteration in order to evaluate convergence behavior.

src/spectrum.jl
```diff
@@ -26,6 +26,21 @@ end
...
+The nonsymmetric Lanczos algorithm is adapted from Algorithm 6.6 in [Saad2011](https://www-users.cse.umn.edu/~saad/eig_book_2ndEd.pdf).
+The running estimate is updated via a [Wallis-Euler recursion](https://en.wikipedia.org/wiki/Continued_fraction).
+"""
+struct Lanczos{T<:Real,IT<:Int} <: SpectrumSolver
+    tol::T
+    maxiter::IT
+    verbose::IT
+end
+
+Lanczos(; tol = 1e-8, maxiter = 5000, verbose = 0) = Lanczos(tol, maxiter, verbose)
```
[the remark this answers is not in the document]
matteosecli: Done in https://github.com/qutip/QuantumToolbox.jl/pull/476/commits/fce5537231eefb6bcf6a01dd043deefb85cf740b.

src/spectrum.jl
```diff
@@ -144,6 +160,143 @@ function _spectrum(
...
+    #spec = Vector{fT}(undef, Length)
+
+    # Calculate |v₁> = B|ρss>
+    ρss = mat2vec(steadystate(L))
+    vₖ = Array{cT}((spre(B) * ρss).data)
+
+    # Calculate <w₁| = <I|A
+    D = prod(L.dimensions)
+    Ivec = SparseVector(D^2, [1 + n * (D + 1) for n in 0:(D-1)], ones(cT, D)) # same as vec(system_identity_matrix)
+    wₖ = transpose(typeof(vₖ)(Ivec)) * spre(A).data
```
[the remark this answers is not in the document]
ytdHuang: ahh, sorry, should be `_spre(A.data)`

matteosecli:
I should have implemented GPU support for the `Lanczos` solver, but testing it as-is (and as a matter of fact, _any_ solver) doesn't work, and you'd have to change this line:
```julia
    ρss = mat2vec(steadystate(L))
```
to
```julia
    ρss = mat2vec(steadystate(L; solver = SteadyStateLinearSolver()))
```
to make it work (and other solvers like `PseudoInverse` too).

A simple example taken from the README:
```julia
using QuantumToolbox
using CUDA
CUDA.allowscalar(false) # Avoid unexpected scalar indexing

N = 20 # cutoff of the Hilbert space dimension
ω = 1.0 # frequency of the harmonic oscillator
γ = 0.1 # damping rate

a_gpu = cu(destroy(N)) # The only difference in the code is the cu() function

H_gpu = ω * a_gpu' * a_gpu

ψ0_gpu = cu(fock(N, 3))

c_ops = [sqrt(γ) * a_gpu]
e_ops = [a_gpu' * a_gpu]

L = liouvillian(H_gpu, c_ops)

ρss = steadystate(L) # doesn't work
# ρss = steadystate(L, solver = SteadyStateLinearSolver()) # this works
```

I seem to remember that `steadystate` used to work, maybe something broke in the meantime?

src/spectrum.jl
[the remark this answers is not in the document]
matteosecli: Also getting an analogous error 🤔

<details>

```
ERROR: MethodError: no method matching _spre(::LinearAlgebra.Adjoint{ComplexF64, CUDA.CUSPARSE.CuSparseMatrixCSC{ComplexF64, Int32}})
The function `_spre` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  _spre(::AbstractMatrix, ::AbstractMatrix)
   @ QuantumToolbox ~/.julia/dev/QuantumToolbox/src/qobj/superoperators.jl:9
  _spre(::SciMLOperators.AddedOperator, ::AbstractMatrix)
   @ QuantumToolbox ~/.julia/dev/QuantumToolbox/src/qobj/superoperators.jl:27
  _spre(::SciMLOperators.ScaledOperator, ::AbstractMatrix)
   @ QuantumToolbox ~/.julia/dev/QuantumToolbox/src/qobj/superoperators.jl:26
  ...
```

</details>

matteosecli:
Btw maybe it's worth keeping `kwargs` to let the user specify the `steadystate` solver, in this case? I'm not sure about the other solvers, but this is basically effortlessly GPU-compatible (it's just spmat/vec products), as long as the steady-state calculation is also GPU-compatible.

matteosecli:
> Btw maybe it's worth keeping `kwargs` to let the user specify the `steadystate` solver, in this case? I'm not sure about the other solvers, but this is basically effortlessly GPU-compatible (it's just spmat/vec products), as long as the steady-state calculation is also GPU-compatible.

Or, the steady-state solver could be an additional field in the `Lanczos` struct.

src/spectrum.jl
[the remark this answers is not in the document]
albertomercurio: I think this is a known issue. Try to do `_spre(sparse(A.data))`.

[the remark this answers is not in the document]
matteosecli: > I think this is a known issue. Try to do `_spre(sparse(A.data))`.

I get:

<details>

```console
ERROR: MethodError: no method matching _spre(::CUDA.CUSPARSE.CuSparseMatrixCSC{ComplexF64, Int32})
The function `_spre` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  _spre(::SparseArrays.AbstractSparseMatrix, ::AbstractMatrix)
   @ QuantumToolbox ~/.julia/dev/QuantumToolbox/src/qobj/superoperators.jl:10
  _spre(::AbstractMatrix, ::AbstractMatrix)
   @ QuantumToolbox ~/.julia/dev/QuantumToolbox/src/qobj/superoperators.jl:9
  _spre(::SciMLOperators.AddedOperator, ::AbstractMatrix)
   @ QuantumToolbox ~/.julia/dev/QuantumToolbox/src/qobj/superoperators.jl:27
  ...
```

</details>

matteosecli:
> > Or, the steady-state solver could be an additional field in the `Lanczos` struct.
> 
> I think this option is better. Better to keep `kwargs` for other options, e.g. for LinearSolve.

I've done that, although I had to modify the include order in `QuantumToolbox.jl` so that the `SteadyStateSolver` abstract type defined in `steadystate.jl` is available to the Lanczos solver in `spectrum.jl`.

With this change, something like
```julia
solver = Lanczos(steadystate_solver = SteadyStateLinearSolver())
spec = spectrum(H, ωlist, c_ops, A, B; solver = solver)
```
now works if `H`, `c_ops`, `A`, and `B` are defined on GPU.

src/spectrum.jl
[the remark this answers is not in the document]
albertomercurio: Yes, because `_spre` requires the identity matrix as extra argument. `spre` should be enough

src/spectrum.jl
```diff
@@ -144,6 +159,142 @@ function _spectrum(
...
+)
+    check_dimensions(L, A, B)
+
+    # Define type shortcuts
+    fT = _FType(L)
+    cT = _CType(L)
+
+    # Calculate |v₁> = B|ρss>
+    ρss =
+        solver.steadystate_solver === nothing ? mat2vec(steadystate(L)) :
```
albertomercurio: `isnothing` should make it more compact

src/spectrum.jl
```diff
@@ -144,6 +159,142 @@ function _spectrum(
...
+        mat2vec(steadystate(L; solver = solver.steadystate_solver))
+    vₖ = (spre(B) * ρss).data
+
+    # Define (possibly GPU) vector type
+    vT = typeof(vₖ)
+
+    # Calculate <w₁| = <I|A
+    D = prod(L.dimensions)
+    Ivec = SparseVector(D^2, [1 + n * (D + 1) for n in 0:(D-1)], ones(cT, D)) # same as vec(system_identity_matrix)
```
albertomercurio, line 186: How is the GPU supported if we have this CPU vector?

src/spectrum.jl
```diff
@@ -144,6 +159,142 @@ function _spectrum(
...
+    w₊₁ = vT(zeros(cT, D^2))'
+
+    # Frequency of renormalization
+    renormFrequency::typeof(solver.maxiter) = 1
+
+    # Loop over the Krylov subspace(s)
+    for k in 1:solver.maxiter
+        # k-th diagonal element
+        w₊₁ = wₖ * L.data
+        αₖ = w₊₁ * vₖ
```
albertomercurio: Don't these two lines allocate memory at each iteration?

src/spectrum.jl
```diff
@@ -144,6 +159,142 @@ function _spectrum(
...
+        if maxResidue <= solver.tol
+            if solver.verbose > 1
+                println("spectrum(): solver::Lanczos converged after $(k) iterations")
+            end
+            break
+        end
+
+        # (k+1)-th left/right vectors, orthogonal to previous ones
+        # Consider using explicit BLAS calls
+        v₊₁ = L.data * vₖ
```
albertomercurio: Same here. Doesn't it allocate?

src/spectrum.jl
albertomercurio: I was thinking that, instead of making `steadystate_solver` `Nothing` as default, one could choose it as the direct one as default (as in steady state), and then avoid using this if condition

src/spectrum.jl
```diff
@@ -144,6 +159,142 @@ function _spectrum(
...
+    maxResidue = fT(0.0)
+
+    # Previous and next left/right Krylov vectors
+    v₋₁ = vT(zeros(cT, D^2))
+    v₊₁ = vT(zeros(cT, D^2))
+    w₋₁ = vT(zeros(cT, D^2))'
+    w₊₁ = vT(zeros(cT, D^2))'
+
+    # Frequency of renormalization
+    renormFrequency::typeof(solver.maxiter) = 1
```
albertomercurio: We do we need to specify the type here?

src/spectrum.jl
matteosecli: > I was thinking that, instead of making `steadystate_solver` `Nothing` as default, one could choose it as the direct one as default (as in steady state), and then avoid using this if condition

I wanted to provide a more predictable and uniform behavior for the user, since `nothing` will use whatever the default solver is in `spectrum`; changing that default will also change this one, so maybe it's easier and one doesn't have to track default solvers across different functions. But as you prefer, really; what's @ytdHuang preference on this?

src/spectrum.jl
matteosecli: I recast as `vT()` in the following line, so it gets transferred to GPU if `vT` is a GPU type.

src/spectrum.jl
matteosecli: So that comparison does not occur accidentally across different types.

src/spectrum.jl
matteosecli: Likely yes, I can use explicit BLAS calls throughout.

src/spectrum.jl
matteosecli: Ditto.

src/spectrum.jl
matteosecli: Changed to default integer

src/spectrum.jl
matteosecli: Should be solved.

src/spectrum.jl
matteosecli: Should be solved.

albertomercurio:
Could you add the changelogs? Remember to do `make changelog` then

ytdHuang:
Somehow I think the previous two comments (merging from the `main` branch) are not done in a correct way...

Now those changes are overwrite again in this PR...

matteosecli:
> Somehow I think the previous two comments (merging from the `main` branch) are not done in a correct way...
> 
> Those changes are now overwritten again in this PR...

Hi @ytdHuang, do you get a merge conflict or something similar in GitHub’s interface? I’ve just rebased my branch against the upstream changes, rather than merging the changes in a separate commit. This was to ensure that my changes were compatible with the updated upstream, especially regarding tests. Is there something missing?

ytdHuang:
After you force-pushed it, now it looks fine !

ytdHuang (approved):

PR #479, opened 2025-06-04
Update index.html

matteosecli:
Fix non-uniform data series

## Checklist

## Description
Fixes data visualization for non-uniform data series (in case, e.g., a new feature is added to benchmarks).

### DIFF
--- a/benchmarks/index.html
+++ b/benchmarks/index.html
@@ -224,6 +224,14 @@
             canvas.width = 10;
             parent.appendChild(canvas);
 
+            const commitIdSet = new Set();
+            for (const benches of dataset.values()) {
+              for (const d of benches) {
+                commitIdSet.add(d.commit.id);
+              }
+            }
+            const commitIds = Array.from(commitIdSet);
+
             const op_to_color = new Map();
             var idx = 0;
             const datasets = [];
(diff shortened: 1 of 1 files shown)


benchmarks/index.html
```diff
@@ -262,12 +273,22 @@ <h1>Benchmarks</h1>
                     afterTitle: items => {
                       const label = items[0].dataset.label;
                       const index = items[0].dataIndex;
-                      const data = dataset.get(label)[index];
+                      const commitId = commitIds[index];
+                      const dataArr = dataset.get(label);
+                      if (!dataArr) return '';
+                      const data = dataArr.find(d => d.commit.id === commitId);
+                      if (!data) return '';
```
albertomercurio, line 280: Why do we need these lines?

matteosecli: Initially, without these lines (and the ones below), the `Lanczos` series datapoint was showing correctly in terms of position along the axis, but the tooltip wasn't showing up. The reason is that it was not correctly matching the commit ID, because, say, index 1 in the `Lanczos` series has not the same commit as index 1 of the `PseudoInverse` series. So we take the correct commit ID from `commitIds`, and we look for the correct benchmark entry by matching its commit ID.

matteosecli: Although I'll admit that rather than opening the correct commit when you click on a datapoint, it opens the first one in the same series. It was doing that even before these changes, though I didn't fix that.

matteosecli: > Although I'll admit that rather than opening the correct commit when you click on a datapoint, it opens the first one in the same series. It was doing that even before these changes, though I didn't fix that.

Alright, the latest commit should fix this, too.

PR #486, opened 2025-06-08
Implement `Base.copy` for `AbstractQuantumObject`

ytdHuang:
## Checklist

## Description
As title

### DIFF
--- a/src/qobj/quantum_object_base.jl
+++ b/src/qobj/quantum_object_base.jl
@@ -1,6 +1,6 @@
 #=  
 This file defines the AbstractQuantumObject structure, all the type structures for AbstractQuantumObject, and fundamental functions in Julia standard library:  
-    - Base: show, length, size, eltype, getindex, setindex!, isequal, :(==), isapprox  
+    - Base: show, length, size, copy, eltype, getindex, setindex!, isequal, :(==), isapprox  
 =#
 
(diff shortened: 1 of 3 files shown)


src/qobj/quantum_object_base.jl
```diff
@@ -93,6 +93,8 @@ Optionally, you can specify an index (`idx`) to just get the corresponding dimen
 Base.size(A::AbstractQuantumObject) = size(A.data)
 Base.size(A::AbstractQuantumObject, idx::Int) = size(A.data, idx)
 
+Base.copy(A::AbstractQuantumObject) = get_typename_wrapper(A)(copy(A.data), A.type, A.dimensions)
```
albertomercurio, line 96: Do you think we need also to copy `dimensions`? They may have in general an internal structure to be copied

ytdHuang: Hmm...
I think copy a quantum object should still preserve the dimension info. since the `size` of the `data` are still the same.

User should not change the dimensions by accessing it directly. Instead, they should create a new dimension object and create the `Qobj` again ?

albertomercurio: Yes, just be sure that in the future we don't include something mutable inside (like a vector)

PR #509, opened 2025-07-23
Improve efficiency of generation of Bloch-Redfield tensor and fix documentation typos

albertomercurio:
## Checklist

## Description
The current implementation of the `bloch_redfield_tensor` function rotates the basis at the end. This is very inefficient as the rotation superoperator is a large dense matrix.

In this PR, I rotate the basis (if asked) directly during the computation. Avoiding to define the rotation operator. This surely improves the `fock_basis=false` case, and partially improves the true case.

### DIFF
--- a/docs/src/users_guide/settings.md
+++ b/docs/src/users_guide/settings.md
@@ -13,7 +13,7 @@ Here, we list out each setting along with the specific functions that will use i
 
 - `tidyup_tol::Float64 = 1e-14` : tolerance for [`tidyup`](@ref) and [`tidyup!`](@ref).
 - `auto_tidyup::Bool = true` : Automatically tidyup during the following situations:
-    * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), and [`eigsolve_al`](@ref).
+    * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), [`eigsolve_al`](@ref), [`bloch_redfield_tensor`](@ref), [`brterm`](@ref) and [`brmesolve`](@ref).
 - (to be announced)
 
 ## Change default settings
--- a/docs/src/users_guide/settings.md
+++ b/docs/src/users_guide/settings.md
@@ -13,7 +13,8 @@ Here, we list out each setting along with the specific functions that will use i
 
 - `tidyup_tol::Float64 = 1e-14` : tolerance for [`tidyup`](@ref) and [`tidyup!`](@ref).
 - `auto_tidyup::Bool = true` : Automatically tidyup during the following situations:
-    * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), [`eigsolve_al`](@ref), [`bloch_redfield_tensor`](@ref), [`brterm`](@ref) and [`brmesolve`](@ref).
+    * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), [`eigsolve_al`](@ref)
+    * Creating [`bloch_redfield_tensor`](@ref) or [`brterm`](@ref), and solving [`brmesolve`](@ref).
 - (to be announced)
 
 ## Change default settings
--- a/docs/src/users_guide/time_evolution/brmesolve.md
+++ b/docs/src/users_guide/time_evolution/brmesolve.md
@@ -162,7 +162,7 @@ H = -Δ/2.0 * sigmax() - ε0/2 * sigmaz()
 
 ohmic_spectrum(ω) = (ω == 0.0) ? γ1 : γ1 / 2 * (ω / (2 * π)) * (ω > 0.0)
 
-R, U = bloch_redfield_tensor(H, [(sigmax(), ohmic_spectrum)])
+R, U = bloch_redfield_tensor(H, ((sigmax(), ohmic_spectrum), ))
 
 R
 ```
@@ -183,6 +183,8 @@ H = - Δ/2.0 * sigmax() - ϵ0/2.0 * sigmaz()
 ohmic_spectrum(ω) = (ω == 0.0) ? γ1 : γ1 / 2 * (ω / (2 * π)) * (ω > 0.0)
 
 a_ops = ((sigmax(), ohmic_spectrum),)
+R = bloch_redfield_tensor(H, a_ops; fock_basis = Val(true))
+
 e_ops = [sigmax(), sigmay(), sigmaz()]
 
 # same initial random ket state in QuTiP doc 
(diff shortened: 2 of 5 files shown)


src/time_evolution/brmesolve.jl
```diff
@@ -114,6 +112,13 @@ function _brterm(
     ac_term = (A_mat .* spectrum) * A_mat
     bd_term = A_mat * (A_mat .* trans(spectrum))
 
+    # Remove small values before passing in the Liouville space
+    if settings.auto_tidyup
```
ytdHuang: I remember that we list out all the situations we will apply `auto_tidyup` in docs here https://qutip.org/QuantumToolbox.jl/stable/users_guide/settings#List-of-settings.

Can you also add `bloch_redfield_tensor` and `brterm` into the list ?

src/time_evolution/brmesolve.jl
```diff
@@ -37,18 +37,16 @@ function bloch_redfield_tensor(
...
-    # in fock basis
-    R0 = liouvillian(H, c_ops)
+    H_new = getVal(fock_basis) ? H : QuantumObject(Diagonal(real.(rst.values)), Operator(), H.dimensions)
+    c_ops_new = isnothing(c_ops) ? nothing : map(x -> getVal(fock_basis) ? x : U' * x * U, c_ops)
+    R = liouvillian(H_new, c_ops_new)
 
-    # set fock_basis=Val(false) and change basis together at the end
-    R1 = 0
-    isempty(a_ops) || (R1 += mapreduce(x -> _brterm(rst, x[1], x[2], sec_cutoff, Val(false)), +, a_ops))
+    isempty(a_ops) || (R += sum(x -> _brterm(rst, x[1], x[2], sec_cutoff, fock_basis), a_ops))
```
ytdHuang: I'm not sure of all the changes from this part.

The original tricky way is that we set `_brterm(...; fock_basis = Val(false))`, cause we can just transform back to Fock basis at the end.

Otherwise `fock_basis = Val(true)` in `_brterm` will still create `SU = sprepost(U, U')`, which is very inefficient when we have more than one `a_op`.

Somehow I think the original way is more efficient ?

Or maybe we can keep the `H_new`, `c_ops_new`, and `R` above, and handle the `a_ops` basis transformation directly in the final `if-else-return`, namely

```julia
H_new = ...
c_ops_new = ...
R = ...
# (the above are same as your implementation)

R_a_ops = 0
isempty(a_ops) || (R_a_ops += sum(x -> _brterm(rst, x[1], x[2], sec_cutoff, Val(false)), a_ops)) # directly set fock_basis to Val(false)

if getVal(fock_basis)
    SU = sprepost(U, U') # transformation matrix from eigen basis back to fock basis
    return R + SU * R_a_ops * SU'
else
    return R + R_a_ops, U
end
```

src/time_evolution/brmesolve.jl
```diff
@@ -37,18 +37,16 @@ function bloch_redfield_tensor(
     U = QuantumObject(rst.vectors, Operator(), H.dimensions)
     sec_cutoff = float(sec_cutoff)
 
-    # in fock basis
-    R0 = liouvillian(H, c_ops)
+    H_new = getVal(fock_basis) ? H : QuantumObject(Diagonal(real.(rst.values)), Operator(), H.dimensions)
```
ytdHuang:
```suggestion
    H_new = getVal(fock_basis) ? H : QuantumObject(Diagonal(rst.values), Operator(), H.dimensions)
```

I suggest we keep the imaginary parts of `Hamiltonian` eigenvalues. Just in case if someone wants to consider non-Hermitian Hamiltonian

src/time_evolution/brmesolve.jl
ytdHuang: BTW, should we `tidyup` here, or after the `if sec_cutoff != -1`-block ?

docs/src/users_guide/time_evolution/brmesolve.md
```diff
@@ -192,7 +194,7 @@ e_ops = [sigmax(), sigmay(), sigmaz()]
 ])
 
 tlist = LinRange(0, 15.0, 1000)
-sol = brmesolve(H, ψ0, tlist, a_ops, e_ops=e_ops)
+sol = mesolve(R, ψ0, tlist, e_ops=e_ops)
```
ytdHuang, line 197: if we do this update, then the entire docs does not demonstrate `brmesolve`. Can you add a few lines (maybe a few words in the plain text above and a commented code here) saying that `brmesolve` is an alternative way to automatically construct `bloch_redfield_tensor` and solve with `mesolve` ?

albertomercurio: Isn't it explained just below?

src/time_evolution/brmesolve.jl
albertomercurio: This is exaclty the core change of this PR. In the default case of `fock_basis=Val(false)`, `SU = sprepost(U, U')` is never created. Being `SU` a huge dense matrix, this is very inefficient.

I also have some ideas on directly removing the definition of this huge dense matrix, and directly change basis in the `a_ops` (if `fock_basis=Val(true)`) of course. But this can be done in the future.

src/time_evolution/brmesolve.jl
albertomercurio: Yeah correct.

docs/src/users_guide/time_evolution/brmesolve.md
ytdHuang: Ok, I just saw it.

docs/src/users_guide/settings.md
```diff
@@ -13,7 +13,7 @@ Here, we list out each setting along with the specific functions that will use i
 
 - `tidyup_tol::Float64 = 1e-14` : tolerance for [`tidyup`](@ref) and [`tidyup!`](@ref).
 - `auto_tidyup::Bool = true` : Automatically tidyup during the following situations:
-    * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), and [`eigsolve_al`](@ref).
+    * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), [`eigsolve_al`](@ref), [`bloch_redfield_tensor`](@ref), [`brterm`](@ref) and [`brmesolve`](@ref).
```
ytdHuang:
```suggestion
    * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), [`eigsolve_al`](@ref)
    * Creating [`bloch_redfield_tensor`](@ref) or [`brterm`](@ref), and solving [`brmesolve`](@ref).
```

We can make it as two items (sub-list)

src/time_evolution/brmesolve.jl
ytdHuang: @albertomercurio 

I suggest we do the change here or just use the code I suggested above.

Because most of the case, user solves the evolution. And `brmesolve` always uses `fock_basis = Val(true)`

albertomercurio: Ok I managed to change basis directly in the Hamiltonian space when `sec_cutoff=-1`. If not, then we still need the `SU` operator to transform the entire Liouvillian after the action of `M_cut`.

PR #515, opened 2025-07-27
Add support to ForwarDiff differentiation

albertomercurio:
## Description
This PR adds the support for autodiff with ForwardDiff.jl. Indeed, it was just missing a custom method for the `_CType` function. 

Moreover, I believe that the previous name of the `_CType` function was not 100% clear, and also unusual for a name of a Julia function. So, here I have also change `_CType` and `_FType` to `_complex_float_type` and `_float_type` respectively.


ytdHuang:
Can we revert to `_FType` and `_CType` ... ?

I use quite a lot in `HierarchicalEOM.jl`...

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `average_states`
     - `average_expect`
     - `std_expect`
+- Add support to ForwardDiff.jl for `sesolve` and `mesolve`. ([#515])
```
ytdHuang, line 22: Since the `keep_runs_results` log has many contents, I suggest we put this one before it.

This should increase the readability of the change log

ytdHuang (approved):

PR #609, opened 2025-12-10
Add support to dense `eigen` on GPU

albertomercurio:
## Checklist

## Description

The support to dense `eigen` solvers was missing. The simple case `eigenstates(H_gpu)` was not working due to missing methods for `to_dense(::Type{CuSparseMatrix...})`. That method was called inside the `eigen` function in order to force type stability on CPU (see https://github.com/JuliaLang/LinearAlgebra.jl/issues/1498). However, this just gives a weak `Union` type on Julia > v1.10, which is not a huge problem. However, this creates a type instability in Julia v1.10, but I think it can be kept like this since it is a problem of that version. So I have decided to remove this forcing type conversion and keep the code simpler.

Since this now gives a `Union` type, while the sparse solver doesn't, I have decided to use `Val` for the `sparse` keyword argument, such that the `sparse = Val(true)` case doesn't return the `Union` type.

### DIFF
--- a/test/ext-test/gpu/Project.toml
+++ b/test/ext-test/gpu/Project.toml
@@ -4,8 +4,5 @@ CUDSS = "45b445bb-4962-46a0-9369-b4df9d0f772e"
 LinearSolve = "7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"
 QuantumToolbox = "6c2fb7c5-b903-41d2-bc5e-5a7c320b9fab"
 
-[sources]
-QuantumToolbox = {path = "/home/alberto/.julia/dev/QuantumToolbox"}
-
 [compat]
 CUDA = "5"


test/ext-test/gpu/Project.toml
```diff
@@ -4,5 +4,8 @@ CUDSS = "45b445bb-4962-46a0-9369-b4df9d0f772e"
 LinearSolve = "7ed4a6bd-45f5-4d41-b270-4a48e9bafcae"
 QuantumToolbox = "6c2fb7c5-b903-41d2-bc5e-5a7c320b9fab"
 
+[sources]
+QuantumToolbox = {path = "/home/alberto/.julia/dev/QuantumToolbox"}
```
ytdHuang: I think this should be removed ?

albertomercurio: Yes indeed.

ytdHuang (approved):

PR #613, opened 2025-12-15
Optimize and fix thermal_dm for nbar=0

yuyichao:
## Checklist

- [x] Please read [Contributing to Quantum Toolbox in Julia](https://qutip.org/QuantumToolbox.jl/stable/resources/contributing).
- [x] Any code changes were done in a way that does not break public API.
- [x] Appropriate tests were added and tested locally by running: `make test`.
- [x] Any code changes should be `julia` formatted by running: `make format`.
- [x] All documents (in `docs/` folder) related to code changes were updated and able to build locally by running: `make docs`.
- [x] (If necessary) the `CHANGELOG.md` should be updated (regarding to the code changes) and built by running: `make changelog`.

## Description
Fix `thermal_dm` for `nbar=0`, which previous returns `NaN` results.

```julia
julia> thermal_dm(3, 0.0)

Quantum Object:   type=Operator()   dims=[3]   size=(3, 3)   ishermitian=false
3×3 Matrix{ComplexF64}:
 NaN+NaN*im  0.0+0.0im   0.0+0.0im
 0.0+0.0im   NaN+NaN*im  0.0+0.0im
 0.0+0.0im   0.0+0.0im   NaN+NaN*im
```

The main fix is to avoid evaluating `exp(-β * 0)` when filling the diagonal data array. Other changes that improves performances includes
* pre-computing the normalization factor to avoid looping through the array twice
* use precomputed value/specialized expressions for small `j` (this mainly matters for `BigFloat` results and doesn't hurt for machine size numbers)
* Use a loop directly since the syntax of using branches in comprehension, while allowed and totally works, is not as elegant anymore. It also reduces reliance on inlining a little bit.

Here's the measured performance difference compared to master. For dense matrix at larger size the performance is going to be dominated by filling the matrix but for most other cases the performance difference is ~10-30%.

```diff
--- master      2025-12-15 17:55:40.759069248 -0500
+++ pr  2025-12-15 17:55:50.030156180 -0500
@@ -1,35 +1,35 @@
 julia> @btime thermal_dm($(3), $(0.2));
-  97.823 ns (6 allocations: 448 bytes)
+  62.755 ns (4 allocations: 336 bytes)

 julia> @btime thermal_dm($(30), $(0.2));
-  824.015 ns (7 allocations: 15.27 KiB)
+  738.646 ns (5 allocations: 14.71 KiB)

 julia> @btime thermal_dm($(300), $(0.2));
-  33.833 μs (9 allocations: 1.38 MiB)
[...]

PR #614, opened 2025-12-16
Fix `thermal_dm` and `enr_thermal_dm` for extreme cases

ytdHuang:
## Checklist

## Description

This is an extended fix for PR #613 .

### DIFF
--- a/test/ext-test/cpu/autodiff/autodiff.jl
+++ b/test/ext-test/cpu/autodiff/autodiff.jl
@@ -85,7 +85,6 @@ end
 n_ss(Δ, F, γ) = abs2(F / (Δ + 1im * γ / 2))
 
 @testset "Autodiff" verbose=true begin
-
     @testset "ForwardDiff for thermal_dm" begin
         N = 100  # dimension of the system
         N_op = num(N) # number operator


src/qobj/states.jl
```diff
@@ -118,13 +118,18 @@ Density matrix for a thermal state (generating thermal state probabilities) with
...
-    β = log(1.0 / n + 1.0)
-    data = [exp(-β * ComplexF64(j)) for j in 0:(N-1)]
-    if getVal(sparse)
-        return QuantumObject(spdiagm(0 => data ./ sum(data)), Operator(), N)
+function thermal_dm(N::Int, n::T; sparse::Union{Bool,Val} = Val(false)) where {T<:Real}
+    if n == 0
+        return fock_dm(N, 0; sparse)
     else
-        return QuantumObject(diagm(0 => data ./ sum(data)), Operator(), N)
+        β = log(1.0 / n + 1.0)
```
yuyichao: beta can be infinite even for `n != 0`

```julia
julia> log(1 / nextfloat(0.0) + 1.0)
Inf
```

ytdHuang: When `n = 0`, it will just return `fock_dm(N, 0)`

src/qobj/states.jl
```diff
@@ -118,13 +118,18 @@ Density matrix for a thermal state (generating thermal state probabilities) with
 !!! warning "Beware of type-stability!"
     If you want to keep type stability, it is recommended to use `thermal_dm(N, n, sparse=Val(sparse))` instead of `thermal_dm(N, n, sparse=sparse)`. See [this link](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-value-type) and the [related Section](@ref doc:Type-Stability) about type stability for more details.
 """
-function thermal_dm(N::Int, n::Real; sparse::Union{Bool,Val} = Val(false))
-    β = log(1.0 / n + 1.0)
-    data = [exp(-β * ComplexF64(j)) for j in 0:(N-1)]
-    if getVal(sparse)
-        return QuantumObject(spdiagm(0 => data ./ sum(data)), Operator(), N)
+function thermal_dm(N::Int, n::T; sparse::Union{Bool,Val} = Val(false)) where {T<:Real}
+    if n == 0
```
yuyichao: This branch should be on beta instead of `n`. I originally had that branch but decided to merge it in the normal one to be safer on the type stability and not bother about a special case performance that much.

src/qobj/states.jl
```diff
@@ -118,13 +118,18 @@ Density matrix for a thermal state (generating thermal state probabilities) with
...
+        return fock_dm(N, 0; sparse)
     else
-        return QuantumObject(diagm(0 => data ./ sum(data)), Operator(), N)
+        β = log(1.0 / n + 1.0)
+        P_n = _complex_float_type(T)[exp(-β * j) for j in 0:(N-1)]
+        P_n /= sum(P_n)
```
yuyichao:
```suggestion
        P_n ./= sum(P_n)
```

src/qobj/states.jl
yuyichao: No I mean this version will still return the wrong result for small but nonzero `n`


```julia
julia> thermal_dm(3, nextfloat(0.0))

Quantum Object:   type=Operator()   dims=[3]   size=(3, 3)   ishermitian=false
3×3 Matrix{ComplexF64}:
 NaN+NaN*im  0.0+0.0im   0.0+0.0im
 0.0+0.0im   NaN+NaN*im  0.0+0.0im
 0.0+0.0im   0.0+0.0im   NaN+NaN*im

julia> thermal_dm(3, 0.0)

Quantum Object:   type=Operator()   dims=[3]   size=(3, 3)   ishermitian=true
3×3 Matrix{ComplexF64}:
 1.0+0.0im  0.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im  0.0+0.0im
```

src/qobj/states.jl
```diff
@@ -118,13 +118,18 @@ Density matrix for a thermal state (generating thermal state probabilities) with
...
-    data = [exp(-β * ComplexF64(j)) for j in 0:(N-1)]
-    if getVal(sparse)
-        return QuantumObject(spdiagm(0 => data ./ sum(data)), Operator(), N)
+function thermal_dm(N::Int, n::T; sparse::Union{Bool,Val} = Val(false)) where {T<:Real}
+    if n == 0
+        return fock_dm(N, 0; sparse)
     else
-        return QuantumObject(diagm(0 => data ./ sum(data)), Operator(), N)
+        β = log(1.0 / n + 1.0)
+        P_n = _complex_float_type(T)[exp(-β * j) for j in 0:(N-1)]
```
yuyichao: Another case I've checked is that `complex(T)` works with other random number types like `ForwardDiff.Dual`. Not sure how important it is but maybe add a fallback to `complex(float(T))` ?

src/qobj/states.jl
```diff
@@ -118,13 +118,18 @@ Density matrix for a thermal state (generating thermal state probabilities) with
...
     If you want to keep type stability, it is recommended to use `thermal_dm(N, n, sparse=Val(sparse))` instead of `thermal_dm(N, n, sparse=sparse)`. See [this link](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-value-type) and the [related Section](@ref doc:Type-Stability) about type stability for more details.
 """
-function thermal_dm(N::Int, n::Real; sparse::Union{Bool,Val} = Val(false))
-    β = log(1.0 / n + 1.0)
-    data = [exp(-β * ComplexF64(j)) for j in 0:(N-1)]
-    if getVal(sparse)
-        return QuantumObject(spdiagm(0 => data ./ sum(data)), Operator(), N)
+function thermal_dm(N::Int, n::T; sparse::Union{Bool,Val} = Val(false)) where {T<:Real}
+    if n == 0
+        return fock_dm(N, 0; sparse)
```
yuyichao:
```suggestion
        return fock_dm(N, zero(T); sparse)
```

or something along this line to fix type stability for non-Float64.

src/qobj/states.jl
ytdHuang: I think our `_complex_float_type` supports `ForwardDiff.Dual`

src/qobj/states.jl
ytdHuang: for `fock_dm`, `0` must be integer, it represents the index of basis

yuyichao: OK then there's a type stability issue.

src/qobj/states.jl
yuyichao: Does it?

```julia
julia> QuantumToolbox._complex_float_type(ForwardDiff.Dual{Nothing,Float64,10})
ERROR: MethodError: no method matching _complex_float_type(::Type{ForwardDiff.Dual{Nothing, Float64, 10}})
The function `_complex_float_type` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  _complex_float_type(::Type{Float32})
   @ QuantumToolbox /usr/share/julia/stdlib/v1.14/QuantumToolbox/src/utilities.jl:191
  _complex_float_type(::Type{Float64})
   @ QuantumToolbox /usr/share/julia/stdlib/v1.14/QuantumToolbox/src/utilities.jl:192
  _complex_float_type(::Type{Int32})
   @ QuantumToolbox /usr/share/julia/stdlib/v1.14/QuantumToolbox/src/utilities.jl:189
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[2]:1
```

src/qobj/states.jl
yuyichao: Was this closed because you have locally made the fix or something?

src/qobj/states.jl
```diff
@@ -118,13 +118,14 @@ Density matrix for a thermal state (generating thermal state probabilities) with
 !!! warning "Beware of type-stability!"
     If you want to keep type stability, it is recommended to use `thermal_dm(N, n, sparse=Val(sparse))` instead of `thermal_dm(N, n, sparse=sparse)`. See [this link](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-value-type) and the [related Section](@ref doc:Type-Stability) about type stability for more details.
 """
-function thermal_dm(N::Int, n::Real; sparse::Union{Bool,Val} = Val(false))
-    β = log(1.0 / n + 1.0)
-    data = [exp(-β * ComplexF64(j)) for j in 0:(N-1)]
+function thermal_dm(N::Int, n::T; sparse::Union{Bool,Val} = Val(false)) where {T<:Real}
+    exp_minus_β = @. 1 - (1 / (n + 1)) # use "1-1/(n+1)" instead of "n/(n+1)" for numerical stability
+    P_n = _complex_float_type(T)[exp_minus_β^j for j in 0:(N-1)]
```
yuyichao: Unfortunately this is way slower than the exp version. `thermal_dm(3000, 0.2; sparse=Val(true))` becomes 50% slower than before.

ytdHuang: But this implementation gives correct results for all known extreme cases. I think this is more important.

yuyichao: And that's why I was talking about branching on the finiteness of beta. Since precomputing the normalization factor is problematic for dealing with `nbar=Inf`, simply replacing the first element in the original array already works and is faster, i.e.

```julia
    data =     P_n = _complex_float_type(T)[j == 0 && !isfinite(β) ? one(β) : exp(-β * j) for j in 0:(N-1)]
```

This is doing basically the same thing as the slower version since the `pow` function fallbacks to log and exp.

src/qobj/states.jl
```diff
@@ -118,13 +118,14 @@ Density matrix for a thermal state (generating thermal state probabilities) with
 !!! warning "Beware of type-stability!"
     If you want to keep type stability, it is recommended to use `thermal_dm(N, n, sparse=Val(sparse))` instead of `thermal_dm(N, n, sparse=sparse)`. See [this link](https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-value-type) and the [related Section](@ref doc:Type-Stability) about type stability for more details.
 """
-function thermal_dm(N::Int, n::Real; sparse::Union{Bool,Val} = Val(false))
-    β = log(1.0 / n + 1.0)
-    data = [exp(-β * ComplexF64(j)) for j in 0:(N-1)]
+function thermal_dm(N::Int, n::T; sparse::Union{Bool,Val} = Val(false)) where {T<:Real}
+    exp_minus_β = 1 - (1 / (n + 1)) # use "1-1/(n+1)" instead of "n/(n+1)" for numerical stability (works for n=0 and Inf)
+    P_n = _complex_float_type(T)[exp_minus_β^j for j in 0:(N-1)]
+    P_n /= sum(P_n)
```
yuyichao: What's wrong with the in-place modification?

yuyichao (changes requested):

src/qobj/states.jl
ytdHuang: JET.jl reports some issues, so I keep it as `P_n /= sum(P_n)` instead of `./=`

src/qobj/states.jl
ytdHuang: Ok, I didn't fully understand your point in the beginning

src/qobj/states.jl
yuyichao: What’s the issue? Why is it recommending allocating a new array?

ytdHuang: It relates to something else in our package

I prefer to keep it as it is for now.

yuyichao: So you mean that this use is perfectly fine?

I still don't understand why such a clear cut and simple optimization is problematic. I literally can't see any reason for the non-inplace version to be better. If there is one, it would be good to state it and maybe mention in a comment.

ytdHuang: Yeah, I also don't understand. It somehow makes the compiler feel like we are making the `Qobj` as a 3D-array.

I even think that the issue reported by `JET.jl` is not a problem from our package.

Anyway, I would prefer to merge this bug fix first, we can improve those things in a future and separate PR if necessary.

yuyichao:
> yes, I've just confirmed in my local repo.

By which you mean it didn't work and had to be fixed.

> Yeah, I also don't understand. It somehow makes the compiler feel like we are making the Qobj as a 3D-array.
> I even think that the issue reported by JET.jl is not a problem from our package.
> Anyway, I would prefer to merge this bug fix first, we can improve those things in a future and separate PR if necessary.

How did you concluded that it's the line I was talking about that's triggering the issue?

I don't know if you had any other local JET runs but from the CI I assume you were talking about https://github.com/qutip/QuantumToolbox.jl/actions/runs/20274818814 and https://github.com/qutip/QuantumToolbox.jl/actions/runs/20274756492. Both of which only shows issue for `thermal_dm` and not `enr_thermal_dm` so it seems that the `./=` change wasn't the issue since both functions had it.

There was also not an attempt to apply only the `./=` changes to see if that indeed was the change that is causing the issue. There's always other unrelated changes that was bundled together so at least on the CI there isn't any proof that this change was the culprit. In fact, and to be more specific, both commits that triggers the issue, 5d52016206e8b8cd973e41a57ba7cb37db144147 and c3cc52959d9e44b1a216b5f1f066e6f243a6e867 contains a unnecessary use of `@.` when computing `exp_minus_β` and none of the succeeded runs contain that line. It should still be correct, which is why I didn't bother to mention previously, thinking it's just the local style, but is probably way more likely to confuse any analyzer than making a change in-place, which is strictly less work for the analyzer.

ytdHuang:
The issue of `./=` raised by `JET.jl` should not relate to this PR. Thus, I prefer to fix it in a separate one.

yuyichao:
There’s simply not an issue and fwiw the previous version attempted to use the broadcast version (it’s doing ./) as well so the inplace one is closer to what the previous one was doing. It is the version you have right now that’s making unnecessary and unrelated change by introducing an array division rather than a broadcast compared to what the code used to do.

And what I was saying was that you have not even tried to make the trivial change, and because of that draw the wrong conclusion that there’s an issue at all.

src/qobj/states.jl
albertomercurio: Have we actually resolved this?

src/qobj/states.jl
albertomercurio: @yuyichao what kind of type stability are you referring to?

src/qobj/states.jl
albertomercurio: In the other PR @yuyichao what saying that perhaps it is better to preallocate the vector and then filling, to avoid comprehension. Is it still true here?

I also would avoid the use of the `Boltzmann` function we have just introduced.

src/qobj/states.jl
albertomercurio: I agree that we should avoid allocating extra vectors unnecessary. We should understand what is going on here.

src/qobj/states.jl
yuyichao: Yes it's fixed by branching on the finiteness of beta rather than the zeroness of nbar. This now relies on exp of a negative number to now be NaN which I think is safe to assume. (Mathematically it's bound so if that's not the case it's a bug of `exp`)

src/qobj/states.jl
yuyichao: The type stability is fixed now by not having a toplevel branch. This version of code used to have a type stability problem because `fock_dm` will (or at least this way of calling it would) only return an eltype of `ComplexF64` and that's not necessarily what the other branch does, e.g. when a Dual number of BigFloat is provided as nbar.

src/qobj/states.jl
yuyichao: The avoiding comprehension part was mainly to not write too much code in the comprehension (for coding style only). It is true that comprehension can basically only be as performant as loop and not better but that's not too important.

The code in the loop (and therefore would've been in comprehension) was more complex in the other PR since I wanted to reuse the result for the summation computation. However, since the analytical expression to calculate the sum had issue with infinite nbar and is therefore disgarded, that's less of a problem (granted `1 + 1/n` is already computed but that has always been the case) and the code in the loop is fairly simple for the current version.

src/qobj/states.jl
albertomercurio: Ok, we should indeed handle `BigFloat`s or `Dual` numbers.

albertomercurio:
It seems that @yuyichao is looking for performance improvement and support to `BigFloat` or `Dual` numbers.

Would the following case be better?

```julia
function thermal_dm(N::Int, n::Tn; sparse::Union{Bool,Val} = Val(false)) where {Tn<:Real}
    (n > 0) || throw(DomainError(n, "The argument n must be positive."))

    T = _complex_float_type(Tn)
    
    # Directly compute exp(-β) = n/(1+n) without logarithm
    # This is more efficient and avoids numerical issues
    # Derived from: β = log(1/n + 1) => exp(-β) = n/(1+n)
    # For n → ∞ (infinite temperature), exp(-β) → 1 (maximally mixed state)
    exp_minus_β = isinf(n) ? one(T) : T(n) / (one(T) + T(n))
    
    # Compute partition function using geometric series formula
    # Z = sum_{j=0}^{N-1} exp(-β*j) = (1 - exp(-β*N)) / (1 - exp(-β))
    Z = if abs(one(real(T)) - real(exp_minus_β)) < eps(Tn)
        T(N)
    else
        (one(T) - exp_minus_β^N) / (one(T) - exp_minus_β)
    end
    
    # Create diagonal elements with normalized probabilities
    scale = one(T) / Z
    data = [exp_minus_β^j * scale for j in 0:(N-1)]
    
    if getVal(sparse)
        return QuantumObject(spdiagm(0 => data), Operator(), N)
    else
        return QuantumObject(diagm(0 => data), Operator(), N)
    end
end
```

It is faster than the `main` branch and it correctly handles `thermal_dm(3, Inf)` or `thermal_dm(3, nextfloat(0.0))`.

It also handles `BigFloat`s, but I haven't tried with `Dual` numbers.

albertomercurio:
> But the current code already supports all different input cases: `0`, `Inf`, `BigFloat`, `nextfloat(0.0)`, `ForwardDiff.Dual`

Yes, what I mean is that, since we also want to optimize. Let's choose the most optimal one. Which one is faster?

src/qobj/states.jl
```diff
@@ -118,15 +118,17 @@ Density matrix for a thermal state (generating thermal state probabilities) with
...
+    P /= sum(P)
     if getVal(sparse)
-        return QuantumObject(spdiagm(0 => data ./ sum(data)), Operator(), N)
+        return QuantumObject(spdiagm(0 => P), Operator(), N)
     else
-        return QuantumObject(diagm(0 => data ./ sum(data)), Operator(), N)
+        return QuantumObject(diagm(0 => P), Operator(), N)
     end
 end
+Boltzmann_weight(β::T, E::Int) where {T<:Real} = (E != 0 || isfinite(β)) ? exp(-β * E) : one(T)
```
albertomercurio: Could you rename is with `boltzmann_weight` or even `_boltzmann_weight`?

src/utilities.jl
```diff
@@ -194,7 +194,8 @@ _complex_float_type(::Type{Complex{Int32}}) = ComplexF32
 _complex_float_type(::Type{Complex{Int64}}) = ComplexF64
 _complex_float_type(::Type{Complex{Float32}}) = ComplexF32
 _complex_float_type(::Type{Complex{Float64}}) = ComplexF64
-_complex_float_type(T::Type{<:Complex}) = T # Allow other untracked Complex types, like ForwardDiff.Dual
+_complex_float_type(T::Type{<:Real}) = Complex{T} # Allow other untracked Complex types, like ForwardDiff.Dual
```
albertomercurio, line 199: Have we tested this case here? I'm not sure doing just `Complex{T}` would work.

src/qobj/states.jl
ytdHuang: ok
I will do it tomorrow.

src/utilities.jl
ytdHuang: I test it by myself and it works.

Do you want me to add ForwardDiff.Dual into thermal_dm test

in this case, we need to add ForwardDiff into test's Project.toml

albertomercurio: We already have it in the autodiff tests. We could add some tests about the derivative of the photon number of a thermal state.

src/qobj/states.jl
ytdHuang: As we did for `Ginibre_ensemble`, I move `_Boltzmann_wight` to `utilities.jl`.

src/utilities.jl
ytdHuang: Done

PR #625, opened 2026-01-04
Support `CUDA v5.9.6`

albertomercurio:


ext/QuantumToolboxCUDAExt.jl
```diff
@@ -97,10 +98,10 @@ function cu(
     return CuSparseMatrixCSC{_convert_eltype_wordsize(eltype(A), word_size)}(A)
 end
 
-QuantumToolbox.to_dense(A::MT) where {MT<:AbstractCuSparseArray} = CuArray(A)
+QuantumToolbox.to_dense(A::MT) where {MT<:AbstractGPUSparseArray} = CuArray(A)
```
ytdHuang: instead of using `GPUArrays.jl` in our `CUDA` extension. Maybe we can create a new `GPUArraysExt` ?

Cause I remember that there is a function called `GPUArrays.dense_array_type`, which will return the corresponding GPU dense array.

And the actual type will be handled by each GPU device backends automatically.

albertomercurio: Ok we can put it directly in the already existing `QuantumToolboxGPUArraysExt.jl` file

ytdHuang: ahh, yes, I forgot that we already have one.

albertomercurio: Ok I was thinking that we can't move it to a GPUArrays extension, since we convert it to `CuArray` which requires CUDA. So I have defined

```julia
const AbstractCuSparseArray = Union{CuSparseVector, CuSparseMatrixCSC, CuSparseMatrixCSR, CuSparseMatrixCOO}
```

and kept the previous definition.

ytdHuang: We don't need to call `CuArray`, we can just use `GPUArrays.dense_array_type`

`CUDA` already made methods to handle this:

See the following example (note that it only works for `CUDA v5.9.6+`):

```julia
using SparseArrays
using GPUArrays: AbstractGPUSparseArray, dense_array_type

to_dense_test(A::MT) where {MT<:AbstractGPUSparseArray} = dense_array_type(MT)(A)

using CUDA, CUDA.CUSPARSE

N = 10
M = cu(sparse(rand(N, N))) # CuSparseMatrixCSC

to_dense_test(M) # CuArray
```

ytdHuang: `GPUArrays.jl` got these new handling functions, as listed in there README file:

image

albertomercurio: Ok done

src/time_evolution/callback_helpers/mesolve_callback_helpers.jl
```diff
@@ -24,9 +24,9 @@ function _save_func_mesolve(u, integrator, e_ops, progr, iter, expvals)
     # The advantage of using this convention is that We don't need
     # to reshape u to make it a matrix, but we reshape the e_ops once.
 
-    ρ = u
-    _expect = op -> dot(op, ρ)
-    @. expvals[:, iter[]] = _expect(e_ops)
+    for (i, op) in enumerate(e_ops)
+        expvals[i, iter[]] = dot(op, integrator.u)
+    end
```
ytdHuang, line 29: should we revert this back to

```julia
ρ = u
_expect = op -> dot(op, ρ)
@. expvals[:, iter[]] = _expect(e_ops)
```

Cause I saw the core tests failing

ext/QuantumToolboxCUDAExt.jl
```diff
@@ -4,7 +4,7 @@ using QuantumToolbox
 using QuantumToolbox: makeVal, getVal
 import QuantumToolbox: _sparse_similar, _convert_eltype_wordsize
 import CUDA: cu, CuArray, allowscalar
-import CUDA.CUSPARSE: CuSparseVector, CuSparseMatrixCSC, CuSparseMatrixCSR, AbstractCuSparseArray
+import CUDA.CUSPARSE: CuSparseVector, CuSparseMatrixCSC, CuSparseMatrixCSR, CuSparseMatrixCOO
```
ytdHuang: Do we need `CuSparseMatrixCOO` ?

Cause there is no call of `CuSparseMatrixCOO` in this file.

src/time_evolution/callback_helpers/mesolve_callback_helpers.jl
albertomercurio: I changed it when debugging, but actually there is no need to define a function and perform broadcasting. It is much easier to do the for loop. SO I kept it.

The tests are failing because CUDA.jl has a weird bug, as shown here https://github.com/JuliaGPU/CUDA.jl/issues/3010

Once this bug will be fixed, the tests should pass.

ext/QuantumToolboxCUDAExt.jl
albertomercurio: Right, I will remove them

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Introduce `matrix_heatmap` and `matrix_histogram` for visualizing operators. ([#621])
 - Allow rectangular eigenvector matrices in `liouvillian_generalized` and expand its documentation with proper citation. Rename it to `liouvillian_dressed_nonsecular`. ([#624])
+- Align with last changes in CUDA.jl. Use `GPUArrays.AbstractGPUSparseArray` instead of `CUDA.CUSPARSE.AbstractCuSparseArray` to support more GPU backends. ([#625])
```
ytdHuang: Can you also mention `v5.9.6+` in the changelog ?

So that in the future, it reminds us which changes we are talking about here.

src/time_evolution/callback_helpers/mesolve_callback_helpers.jl
ytdHuang: @albertomercurio 

The runtests show that results are incorrect and allocation tests fails.

I think it's because of this change, should we revert to the original code ?

albertomercurio: Ok I have fixed it. I was using `integrator.u` instead of just the interpolated `u`.

test/ext-test/gpu/cuda_ext.jl
```diff
@@ -93,40 +93,41 @@
...
+
+    L_cpu64 = liouvillian(H_cpu64, c_ops_cpu64)
+    L_gpu64 = liouvillian(H_gpu64, c_ops_gpu64)
+
+    @test SparseMatrixCSC(L_gpu64.data) ≈ L_cpu64.data
 
-    @test all([isapprox(sol_cpu.expect[i], sol_gpu64.expect[i]) for i in 1:length(tlist)])
-    @test all([isapprox(sol_cpu.expect[i], sol_gpu32.expect[i]; atol = 1.0e-6) for i in 1:length(tlist)])
+    @test all([isapprox(sol_cpu64.expect[i], sol_gpu64.expect[i]) for i in 1:length(tlist)])
+    @test all([isapprox(sol_cpu32.expect[i], sol_gpu32.expect[i]) for i in 1:length(tlist)])
```
ytdHuang: @albertomercurio 

I think this is the final minor comment for this PR.

I noticed that the CUDA tests sometimes fail, sometimes not:

```
CUDA Extension: Test Failed at /var/lib/buildkite-agent/builds/gpuci-7/julialang/quantumtoolbox-dot-jl/test/ext-test/gpu/cuda_ext.jl:130

Expression: all([isapprox(sol_cpu32.expect[i], sol_gpu32.expect[i]) for i = 1:length(tlist)])
```

I think this is because 32-bit doesn't have such a high precision `atol=1.0e-10`.

Should we add the `atol = 1.0e-6` back for 32-bit tests:

```suggestion
    @test all([isapprox(sol_cpu32.expect[i], sol_gpu32.expect[i]; atol = 1.0e-6) for i in 1:length(tlist)])
```

I understand the 32-bit results from CPU and GPU should be the same. But I think it might not be so accurate due to machine precision. Or maybe use `atol = 1.0e-8` instead of the default `1.0e-10` ?

albertomercurio: Ok

PR #628, opened 2026-01-08
Add Copilot instructions

albertomercurio:


ytdHuang (approved):
LGTM!

One minor comment: 
I'm not sure whether we should add this to the change log, cause it doesn't directly relate to the package.

PR #653, opened 2026-01-31
Better handling of Float32 and sparse types for time evolution solvers

albertomercurio:
## Checklist

## Description
By default `Eye(N)` gives a Float64 type identity matrix. This creates problems when dealing with Float32 types, as it promotes them to ComplexF64 when doing `spre`, `spost` or similar.

Also, the `spre`-`spost` functions were manually converting any matrix to sparse. This is suboptimal and doesn't give control. An input Dense matrix should remain dense.

Finally, `abstol` and `reltol` were Float64, giving problems with Float32 objects.

With these changes, the following code works on a Metal GPU

```julia
using GenericSparseArrays
using Metal
using Adapt
using QuantumToolbox

# %%

Adapt.@adapt_structure QuantumObject

GenericSparseArrays.GenericSparseMatrixCSR(A::QuantumObject) = QuantumObject(GenericSparseMatrixCSR(A.data), A.type, A.dimensions)

# %%

N = 30

a_cpu = destroy(ComplexF32, N)

a = adapt(MtlArray, GenericSparseMatrixCSR(a_cpu));

H = a' * a + 0.1f0 * (a + a');
c_ops = [sqrt(0.05f0) * a];

ψ0_cpu = fock(ComplexF32, N, 5)
ψ0 = adapt(MtlArray, ψ0_cpu)

tlist = range(0.0f0, 10.0f0, 100)

sol = mesolve(H, ψ0, tlist, c_ops)
```

### DIFF
--- a/docs/src/users_guide/settings.md
+++ b/docs/src/users_guide/settings.md
@@ -11,11 +11,10 @@ All settings are stored in [`QuantumToolbox.settings`](@ref).
 
 Here, we list out each setting along with the specific functions that will use it.
 
-- `tidyup_tol::Float64 = 1.0e-14` : tolerance for [`tidyup`](@ref) and [`tidyup!`](@ref).
+- `tidyup_tol::Float64 = 1.0e-14` : tolerance for [`tidyup`](@ref), [`tidyup!`](@ref) and [`to_sparse`](@ref).
 - `auto_tidyup::Bool = true` : Automatically tidyup during the following situations:
     * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), [`eigsolve_al`](@ref)
     * Creating [`bloch_redfield_tensor`](@ref) or [`brterm`](@ref), and solving [`brmesolve`](@ref).
-    * Converting to sparse using [`to_sparse`](@ref).
 - `ProgressMeterKWARGS::NamedTuple = (showspeed = true, printed = true)` : Default keyword arguments for progress bar in [`ProgressMeter.jl`](https://github.com/timholy/ProgressMeter.jl). This allows the customization of progress bar.
 - (to be announced)
 
--- a/src/settings.jl
+++ b/src/settings.jl
@@ -21,7 +21,7 @@ Contains all the default global settings of QuantumToolbox.jl.
 
 # List of settings
 
-- `tidyup_tol::Float64 = 1.0e-14` : tolerance for [`tidyup`](@ref) and [`tidyup!`](@ref).
+- `tidyup_tol::Float64 = 1.0e-14` : tolerance for [`tidyup`](@ref), [`tidyup!`](@ref) and [`to_sparse`](@ref).
 - `auto_tidyup::Bool = true` : Automatically tidyup.
 - `ProgressMeterKWARGS::NamedTuple = (showspeed = true, printed = true)` : Default keyword arguments for progress bar in [`ProgressMeter.jl`](https://github.com/timholy/ProgressMeter.jl). This allows the customization of progress bar.
 


test/core-test/brmesolve.jl
```diff
@@ -112,6 +119,6 @@ end
         me = mesolve(H, ψ0, tlist, me_c_ops, e_ops = e_ops, progress_bar = Val(false))
         brme = brmesolve(H, ψ0, tlist, brme_a_ops, brme_c_ops, e_ops = e_ops, progress_bar = Val(false))
 
-        @test all(me.expect .== brme.expect)
+        @test me.expect ≈ brme.expect
```
ytdHuang: Why do we change to `isapprox` ?

In these cases, we are testing that `mesolve` and `BRmesolve` should be exactly the same.

src/qobj/superoperators.jl
```diff
@@ -6,14 +6,11 @@ export spre, spost, sprepost, liouvillian, lindblad_dissipator
...
-_spost(B::AbstractSparseMatrix) = kron(transpose(B), Eye(size(B, 1)))
-_sprepost(A::AbstractMatrix, B::AbstractMatrix) = kron(transpose(sparse(B)), sparse(A))
-_sprepost(A::AbstractMatrix, B::AbstractSparseMatrix) = kron(transpose(B), sparse(A))
-_sprepost(A::AbstractSparseMatrix, B::AbstractMatrix) = kron(transpose(sparse(B)), A)
-_sprepost(A::AbstractSparseMatrix, B::AbstractSparseMatrix) = kron(transpose(B), A)
+_spre(A::AbstractMatrix{T}) where {T} = kron(Eye{T}(size(A, 1)), A)
+_spost(B::AbstractMatrix{T}) where {T} = kron(transpose(B), Eye{T}(size(B, 1)))
+_spost(B::Adjoint{T, <:AbstractMatrix}) where {T} = kron(conj(parent(B)), Eye{T}(size(parent(B), 1))) # avoiding nested Transpose{Adjoint} wrapper
+_sprepost(A::AbstractMatrix, B::AbstractMatrix) = kron(transpose(B), A)
+_sprepost(A::AbstractMatrix, B::Adjoint) = kron(conj(parent(B)), A) # avoiding nested Transpose{Adjoint} wrapper
```
ytdHuang, line 13: Is there a reason that we no longer ensure the output is sparse matrix ?

cause I remember we discussed about this (long time ago) that we make sure we always return `sparse` superoperators, as the comment says above.

src/qobj/functions.jl
```diff
@@ -155,6 +155,9 @@ function to_sparse(A::VT, tol::Real = 1.0e-10) where {VT <: AbstractVector}
     return sparsevec(idxs, vals, length(A))
 end
 
+to_sparse_if_needed(::Val{needed}, A::Union{QuantumObject, AbstractArray}, tol::Real = 1.0e-10) where {needed} =
+    needed ? to_sparse(A, tol) : A
```
ytdHuang, line 159: I'm kind of confused between `tidyup` and the effect of `tol` for `to_sparse`.

They have different `tol`erence, and also `tidyup` handles `real` and `imag` part separately.

Should we just use `tidyup` in `to_sparse` ?
Also, maybe use `settings.auto_tidyup` and default `tidyup_tol` in `to_sparse` ?

src/qobj/superoperators.jl
albertomercurio: I remember we were using `sparse` because if you do `spost(adjoint(a))` this calls `kron(transpose(adoint(a)), I)`, but there is no method for this. So applying `sparse` on `adjoint` returns a sparse and the `kron` between a `Transpose` and `Diagonal` is available.

However, as said in the PR description, there is no need to convert it to sparse. I can in general have any abstract matrix format, and suppose to always have that format.

test/core-test/brmesolve.jl
albertomercurio: Ok

src/qobj/functions.jl
albertomercurio: Yes I can use the defaul tolerance here. However, it is better to not call just `tidyup`, because it will first convert a matrix into sparse and then dropping the small elements. It is better to extract the non-vanishing elements direclty from the dense array.

src/qobj/functions.jl
```diff
@@ -141,20 +141,23 @@ to_dense(::Type{M}) where {M <: AbstractMatrix} = M
 
 Converts a dense QuantumObject to a sparse QuantumObject.
 """
-to_sparse(A::QuantumObject, tol::Real = 1.0e-10) = QuantumObject(to_sparse(A.data, tol), A.type, A.dimensions)
-function to_sparse(A::MT, tol::Real = 1.0e-10) where {MT <: AbstractMatrix}
+to_sparse(A::QuantumObject, tol::Real = settings.tidyup_tol) = QuantumObject(to_sparse(A.data, tol), A.type, A.dimensions)
```
ytdHuang, line 144: I just noticed that the docstring for `to_sparse` doesn't mention argument `tol`. Can you add it here, too?

Also, in our documentation https://qutip.org/QuantumToolbox.jl/stable/users_guide/settings#List-of-settings, we list out the situation where we use `QuantumToolbox.settings`, can you also add `to_sparse` into `tidyup_tol` ?

src/qobj/functions.jl
ytdHuang: ok

test/core-test/brmesolve.jl
```diff
@@ -112,6 +119,6 @@ end
         me = mesolve(H, ψ0, tlist, me_c_ops, e_ops = e_ops, progress_bar = Val(false))
         brme = brmesolve(H, ψ0, tlist, brme_a_ops, brme_c_ops, e_ops = e_ops, progress_bar = Val(false))
 
-        @test all(me.expect .== brme.expect)
+        @test me.expect == brme.expect
```
ytdHuang, line 123: can you also add a comment here: `mesolve and brmesolve should be exactly the same for these cases`

(just in case we forgot in the future)

src/qobj/functions.jl
albertomercurio: Ok

test/core-test/brmesolve.jl
albertomercurio: Ok

docs/src/users_guide/settings.md
```diff
@@ -15,6 +15,7 @@ Here, we list out each setting along with the specific functions that will use i
 - `auto_tidyup::Bool = true` : Automatically tidyup during the following situations:
     * Solving for eigenstates, including [`eigenstates`](@ref), [`eigsolve`](@ref), [`eigsolve_al`](@ref)
     * Creating [`bloch_redfield_tensor`](@ref) or [`brterm`](@ref), and solving [`brmesolve`](@ref).
+    * Converting to sparse using [`to_sparse`](@ref).
 - `ProgressMeterKWARGS::NamedTuple = (showspeed = true, printed = true)` : Default keyword arguments for progress bar in [`ProgressMeter.jl`](https://github.com/timholy/ProgressMeter.jl). This allows the customization of progress bar.
```
ytdHuang, line 18: I suggest we put it in `tidyup_tol` instead of `auto_tidyup`. Cause `to_sparse` doesn't depend on `auto_tidyup`:

```markdown
- `tidyup_tol::Float64 = 1.0e-14` : tolerance for [`tidyup`](@ref), [`tidyup!`](@ref), and [`to_sparse`](@ref).
```

ytdHuang: can you mention `to_sparse` in the docstring of `Settings` ?

albertomercurio: Ok

ytdHuang (approved):

PR #707, opened 2026-05-04
Introduce `SuperOperatorMatrixForm` and add `matrix_form` argument

### LINKED ISSUE #617: Suboptimal and bad performance in mesolve
# Disclaimer

This performance bug report is based on my test on a few generic cases that originate from problems that I need to calculate myself. Currently these cases does not include any decay terms but the conclusion should be similar with some decay terms.

# Short version

The approach of treating the derivative calculation in master equation as a sparse matrix multiplied on a linearized density matrix is sub-optimal and can lead to pretty bad performance in many cases. Keeping the operators in their original matrix form is almost always better. There's also additional optimization opportunities that could improve the performance even further.

# Long version

It seems that when dealing with a master equation, QuantumToolbox.jl always turns the (nxn) density matrix into a (n^2) sized vector and turns the RHS of the master equation (which is linear wrt the density matrix) as a (n^2 x n^2) sparse matrix that operates on the "density vector". While this transformation doesn't really changes the complexity, it can often be slower in practice since the new matrix has more non-zero elements compared to the old one. It's especially bad if the original matrix was dense in which case it'll also lost the highly efficient dense matrix multiplication.

Another trick can also be used to reduce the number of matrix multiplications. This is because although the commutator between H and rho nominally requires two matrix multiplication, the second one is simply the Hermitian conjugate of the first one, this allows one to replace the multiplication with a O(n^2) operation, and this is going to be cheaper than the multiplication as long as the Hamiltonian is fully connected (contains at least ~n non-zero elements).

## Benchmarks

I've tested this with two types of problems, one is a Rabi flopping in a two level system that's coupled to the motion of the atom, the other is a dipole transition between two angular momentum manifolds. In the plot of the result below, they are marked with the matrix size and, for sparse version, the density of the matrix.

The four implementation I've tested are, `QuantumToolbox.jl`, `QuantumOptics.jl`, a **Simple** one where I simply did in place `mul!` of rho and H, and an **Optimized** one where I simplified the first `mul!` replaced the second `mul!` with taking the imaginary part.

The algorithm used by `QuantumOptics.jl` is basically exactly the same as the **Simple** one with the exception that they have a custom implementation for some of the sparse matrix multiplication that sometimes out-perform the stock one.

I've tested on 4 machines I have access to and on one of them tested two julia versions due to a known julia performance issue.

All the numbers shown are times taken for the computation normalized to the `QuantumToolbox.jl` number. So smaller number means good and smaller than one means outperforming `QuantumToolbox.jl`

### Result for the dense version:

Image

The performance is clearly better for simply doing the matrix multiplication. (the outlier for the M1 curve is probably some glitches on the system since that's the only laptop tested). The performance gap is larger for the larger system due to triggering the multi-threaded dense matrix multiplication.

The hermitian conjugate trick is also basically always helping even when facing a multithreaded matrix multiplication. I can see that for really big matrix on large number of cores the single threaded may be slower but that's not a fair comparison and the hermitian conjugate trick is also parallelizable.

### Result for the sparse version:

Image

All the non `QuantumToolbox.jl` implementation on julia >= 1.11 (i.e. everything but the orange line) are affected by https://github.com/JuliaLang/julia/issues/60409 so for performance comparison between the flattening the density matrix or not the orange line should be looked at more.

Essentially the only case where flattening seems to have an advantage is the one with smallest 8x8 matrix. This makes sense since the slight overhead of dealing with matrix and the additional loop in the sparse-dense multiplication matters more when there's only 18 non-zero elements to deal with anyway... Even then, the hermitian conjugate trick still offers the best performance in all cases.

# Dealing with decay operators

As mentioned above, the tests so far only deal with master equation without decay terms. However, a lot of the findings can also be extended to cases with decays as well. Everything applies directly to the non-Hermitian Hamiltonian part so the only part that needs to be treated separately are the `Γ ρ Γ†` terms. For very sparse decay operators, a flattened view (which doesn't necessarily need an actual `n^2 x n^2` sparse matrix but just something that stores the same linear operation information) may actually be advantageous. For denser decay operators though, performing the matrix multiplications as is might still be the way to go.

# Potential action items

I can only speak from the view of what's the implementation that should give the best performance. Some fixes may need significant change to the internals of the package. There are two potential changes.

1. Use direct matrix multiplication instead of flattening the density matrix
2. Use Hermitian conjugate to reduce the number of multiplications.

The two are actually completely independent and one can be implemented without the other. Without looking much deeper into the package internal (as well as `SciMLOperators`) I'm not sure which one is going to be an easier fix (probably the second one though).

# Code for testing and raw numbers from the benchmark

https://github.com/yuyichao/yyc-data/tree/f6b6a14d2c8442efb7a8b8aa10dc11a996a5cbcf/euriqa/calculations/master_benchmark

albertomercurio:
## Checklist

## Description
This package currently supports the superoperator representation in the framework of vectorized density matrix. By doing so, every superoperator is represented as a matrix. However, this can be suboptimal in several cases, especially when the system size increases.

Thus, I implemented here the support for the matrix form representation. In this framework, the density matrix remains a matrix (`Operator()`). The right and left-right action is obtained through the use of SciMLOperators.jl.

The user just needs to set `matrix_form = Val(true)` in order to use this framework. For example
- `mesolve(H, psi0, tlist, c_ops; matrix_form = Val(true))`
- `liouvillian(H, c_ops; matrix_form = Val(true))`
- `liouvillian_dressed_nonsecular(H, fields, T_list; matrix_form = Val(true))`

There is currently this PR (https://github.com/SciML/SciMLOperators.jl/pull/370) in SciMLOperators.jl that improves the cache efficiency for such cases, reducing even more the memory usage. This PR cannot be merged before that PR.

## Transverse Field Ising Model Benchmarks

I benchmark now both the memory and the computational efficiency of `mesolve`. The benchmarks are performed on the GPU NVIDIA 4090.

```julia
using LinearAlgebra
using QuantumToolbox
using QuantumToolbox: makeVal, getVal
using CUDA
using ProgressMeter
using Chairmarks
using CairoMakie

# %%

const Jx = 25
const hz = 50

const Δ = 0.1 # Detuning with respect to the drive
const U = -0.05 # Nonlinearity
const F = 2 # Amplitude of the drive
const nth = 0.2 # Thermal photons

const γ = 1 # Decay rate

function multisite_operator_gpu(args...; to_gpu::Val = Val(false))
    op = multisite_operator(args...)
    !getVal(to_gpu) && return op
    return CUSPARSE.CuSparseMatrixCSR(op)
end

function generate_system(N, ::Val{:ising}, to_gpu::Val)
    dims = ntuple(i -> 2, makeVal(N))
    Hz = hz * sum(i -> multisite_operator_gpu(dims, i => sigmaz(), to_gpu = to_gpu), 1:getVal(N))
    Hxx = Jx * sum(i -> multisite_operator_gpu(dims, i => sigmax(), i + 1 => sigmax(), to_gpu = to_gpu), 1:(getVal(N) - 1))
    H = Hz + Hxx

    # c_ops = [sqrt(γ) * local_op(sigmam(), i, N) for i in 1:getVal(N)]
    c_ops = ntuple(i -> sqrt(γ) * multisite_operator_gpu(dims, i => sigmam(), to_gpu = to_gpu), makeVal(N))

    # e_ops = [local_op(sigmaz(), getVal(N), N)]
    e_ops = ntuple(i -> multisite_operator_gpu(dims, i => sigmaz(), to_gpu = to_gpu), makeVal(N))

    return H, c_ops, e_ops
end

function initial_state(N, ::Val{:ising}, to_gpu::Val)
    state = tensor(ntuple(i -> basis(2, 0), makeVal(N))...)
    return getVal(to_gpu) ? cu(state) : state
end

function quantumtoolbox_mesolve(N, system_type::Val; matrix_form = Val(false), to_gpu::Val = Val(false))
    H, c_ops, e_ops = generate_system(N, system_type, to_gpu)

    tlist = range(0, 10, 100)
    ψ0 = initial_state(N, system_type, to_gpu)

    mesolve(H, ψ0, tlist[1:2], c_ops, e_ops = e_ops, progress_bar = Val(false)) # Warm-up

    benchmark_result =
        @be mesolve($H, $ψ0, $tlist, $c_ops, e_ops = $e_ops, progress_bar = Val(false), matrix_form = $matrix_form).expect

    return sum(s -> s.time, benchmark_result.samples) / length(benchmark_result.samples)
end

function run_benchmarks(::Val{Nmax}, ::Val{model}; matrix_form = Val(false), to_gpu = Val(false)) where {Nmax, model}
    Nvals = ntuple(i -> Val(i), Val(Nmax))
    return @showprogress map(Nvals[2:end]) do N
        quantumtoolbox_mesolve(N, Val(model); matrix_form = matrix_form, to_gpu = to_gpu)
    end
end

function run_summarysize(::Val{Nmax}, ::Val{model}; matrix_form = Val(false)) where {Nmax, model}
    Nvals = ntuple(i -> Val(i), Val(Nmax))
    return map(Nvals[2:end]) do N
        H, c_ops, e_ops = generate_system(N, Val(model), Val(false))
        L = liouvillian(H, c_ops; matrix_form = matrix_form)
        ρ = rand_dm(ntuple(i -> 2, makeVal(N)))
        L_cached = getVal(matrix_form) ? cache_operator(L, ρ) : L
        # L_cached = L
        Base.summarysize(L_cached)
    end
end

# %%

Nmax = Val(12)

summarysize_vec = run_summarysize(Nmax, Val(:ising); matrix_form = Val(false))
summarysize_mat = run_summarysize(Nmax, Val(:ising); matrix_form = Val(true))

benchmarks_vec = run_benchmarks(Nmax, Val(:ising); matrix_form = Val(false), to_gpu = Val(true))
benchmarks_mat = run_benchmarks(Nmax, Val(:ising); matrix_form = Val(true), to_gpu = Val(true))

# %%

fig = Figure()
ax_memory = Axis(fig[1, 1], xlabel = "N", ylabel = "Memory (MB)", yscale = log10, xticks = 2:2:getVal(Nmax))
ax_time = Axis(fig[2, 1], xlabel = "N", ylabel = "Time (s)", yscale = log10, xticks = 2:2:getVal(Nmax))

scatterlines!(ax_memory, 2:getVal(Nmax), collect(summarysize_vec) ./ 1.0e6, label = "Vectorized")
scatterlines!(ax_memory, 2:getVal(Nmax), collect(summarysize_mat) ./ 1.0e6, label = "Matrix")
scatterlines!(ax_time, 2:getVal(Nmax), collect(benchmarks_vec), label = "Vectorized")
scatterlines!(ax_time, 2:getVal(Nmax), collect(benchmarks_mat), label = "Matrix")

axislegend(ax_memory; position = :lt)
axislegend(ax_time; position = :lt)

fig
```

image

# Liouvillian Dressed Nonsecular Benchmarks

I then test the `liouvillian_dressed_nonsecular`, which is known to be poorly sparse.

### CPU case

```julia
using QuantumToolbox
using CUDA
using SciMLOperators
using Adapt
using Chairmarks

# %%

N = 9
ωc1 = 2
ωc2 = 1
ωq = 1
g = 0.6
γ1 = 0.01
γ2 = 0.01

a1 = tensor(destroy(N), qeye(N), qeye(2))
a2 = tensor(qeye(N), destroy(N), qeye(2))
σx = tensor(qeye(N), qeye(N), sigmax())
σz = tensor(qeye(N), qeye(N), sigmaz())

H = ωc1 * a1' * a1 + ωc2 * a2' * a2 + ωq / 2 * σz + g * ((a1 + a1') + (a2 + a2')) * (σx + σz)

fields = (√(γ1 / ωc1) * (a1 + a1'), √(γ2 / ωc2) * (a2 + a2'))
T_list = (0.0, 0.0)

L_gme_vec = liouvillian_dressed_nonsecular(H, fields, T_list)[3]
GC.gc(true)
L_gme_mat = liouvillian_dressed_nonsecular(H, fields, T_list; matrix_form = Val(true))[3]

ρ0 = ket2dm(fock(N * N * 2, 0; dims = (N, N, 2)))

L_gme_mat_cached = cache_operator(L_gme_mat, ρ0)

Base.summarysize(L_gme_vec) / Base.summarysize(L_gme_mat_cached)

# %%

tlist = range(0, 100, length = 100)
e_ops = (a1' * a1, )

@be mesolve($L_gme_vec, $ρ0, $tlist; progress_bar = Val(false), e_ops = $(e_ops))
@be mesolve($L_gme_mat, $ρ0, $tlist; progress_bar = Val(false), e_ops = $(e_ops), matrix_form = Val(true))
```

|                      | **Vectorized** | **Matrix Form** | Ratio |
| -------------------- | -------------- | --------------- | ----- |
| Memory Usage (Mb)    | 4083           | 5.99            | 681   |
| Simulation Time (ms) | 4400           | 480             | 9.1   |

### GPU

```julia

Adapt.adapt_structure(to, x::QuantumObject) = QuantumObject(Adapt.adapt_structure(to, x.data), x.type, x.dimensions)
Adapt.adapt_structure(to, x::QuantumObjectEvolution) = QuantumObjectEvolution(Adapt.adapt_structure(to, x.data), x.type, x.dimensions)
Adapt.adapt_structure(to, x::SciMLOperators.AddedOperator) = SciMLOperators.AddedOperator(Adapt.adapt_structure(to, x.ops))
Adapt.adapt_structure(to, x::SciMLOperators.MatrixOperator) = SciMLOperators.MatrixOperator(to(x.A))
Adapt.adapt_structure(to, x::QuantumToolbox.SpostSuperOperator) = QuantumToolbox.SpostSuperOperator(to(x.R))
Adapt.adapt_structure(to, x::QuantumToolbox.SprePostSuperOperator) = QuantumToolbox.SprePostSuperOperator(to(x.L), to(x.R))

L_gme_vec_gpu = CUSPARSE.CuSparseMatrixCSR(L_gme_vec)
L_gme_mat_gpu = adapt(CUSPARSE.CuSparseMatrixCSR, L_gme_mat)
ρ0_gpu = adapt(CuArray, ρ0)

# %%

e_ops_gpu = (CUSPARSE.CuSparseMatrixCSR(a1' * a1), )

@be mesolve($L_gme_vec_gpu, $ρ0_gpu, $tlist; progress_bar = Val(false), e_ops = $(e_ops_gpu))
@be mesolve($L_gme_mat_gpu, $ρ0_gpu, $tlist; progress_bar = Val(false), e_ops = $(e_ops_gpu), matrix_form = Val(true))
```

|                      | **Vectorized** | **Matrix Form** | Ratio |
| -------------------- | -------------- | --------------- | ----- |
| Simulation Time (ms) | 110            | 39              | 2.8   |

## Related Issues

This PR fixes #617


src/qobj/superoperators_scimloperators.jl
```diff
@@ -0,0 +1,138 @@
...
+    iscached(op) || throw(ArgumentError("The cache for the SprePostSuperOperator must be initialized before multiplication. Use `cache_operator` to initialize the cache."))
+    mul!(op.cache, op.L, u)     # cache = L * u
+    mul!(v, op.cache, op.R)     # v = cache * R = L * u * R
+    return v
+end
+
+function LinearAlgebra.mul!(v::AbstractMatrix, op::SprePostSuperOperator, u::AbstractMatrix, α::Number, β::Number)
+    iscached(op) || throw(ArgumentError("The cache for the SprePostSuperOperator must be initialized before multiplication. Use `cache_operator` to initialize the cache."))
+    mul!(op.cache, op.L, u)            # cache = L * u
+    mul!(v, op.cache, op.R, α, β)     # v = α * (L * u * R) + β * v
```
TendonFFF, line 110: this is potentially dense * sparse, and hurt performance greatly

albertomercurio: Don't we have already have efficient methods for dense * sparse multiplication?

src/qobj/superoperators.jl
```diff
@@ -184,26 +219,53 @@ function liouvillian(
...
+
+        return L
     end
-    return L
 end
 
-liouvillian(H::Nothing, c_ops::Union{AbstractVector, Tuple}; kwargs...) = _sum_lindblad_dissipators(c_ops)
-
+liouvillian(H::Nothing, c_ops::AbstractVecOrTuple; kwargs...) = _sum_lindblad_dissipators(c_ops)
 liouvillian(H::Nothing, c_ops::Nothing; kwargs...) = 0
```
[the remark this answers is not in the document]
albertomercurio: Fixed

ytdHuang:
should we merge PR #751 first?

cause some of these updates should be moved to `Core` library

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [lib] Introduce `QuantumToolboxCore` library. ([#686], [#753])
 - [lib] Move more basic functionalities to `QuantumToolboxCore` library. ([#751])
+- Add support for matrix form `liouvillian`, `liouvillian_dressed_nonsecular` and `mesolve` time evolution. By setting `matrix_form = Val(true)`, the density matrix is not vectorized, which improves the memory efficiency for large systems. The superoperators are now stored through SciMLOperators.jl objects. ([#658])
```
ytdHuang: I thought this PR number is `707` ?

CHANGELOG.md
```diff
@@ -530,6 +531,7 @@ Release date: 2024-11-13
 [#653]: https://github.com/qutip/QuantumToolbox.jl/issues/653
 [#656]: https://github.com/qutip/QuantumToolbox.jl/issues/656
 [#657]: https://github.com/qutip/QuantumToolbox.jl/issues/657
+[#658]: https://github.com/qutip/QuantumToolbox.jl/issues/658
```
ytdHuang: same here, should remove this, and change it to `707`

lib/QuantumToolboxCore/src/qobj/superoperators_scimloperators.jl
```diff
@@ -0,0 +1,138 @@
...
+These operate on density matrices in matrix form (N×N) rather than vectorized (N²) form.
+=#
+
+export SpostSuperOperator, SprePostSuperOperator
+
+@doc raw"""
+    SpostSuperOperator{T, OpType <: Union{AbstractMatrix, AbstractSciMLOperator}} <: AbstractSciMLOperator{T}
+
+Represents the superoperator ``\mathcal{O}(\hat{B})[\hat{\rho}] = \hat{\rho}\hat{B}`` acting on the **non-vectorized**
+density operator matrix ``\hat{\rho}``.
```
ytdHuang: we can just make this one-line ?

and maybe just call it density matrix instead of density operator matrix ?

```suggestion
Represents the superoperator ``\mathcal{O}(\hat{B})[\hat{\rho}] = \hat{\rho}\hat{B}`` acting on the **non-vectorized** density matrix ``\hat{\rho}``.
```

lib/QuantumToolboxCore/src/qobj/superoperators_scimloperators.jl
```diff
@@ -0,0 +1,138 @@
...
+
+The action is defined as:
+```math
+\mathcal{O}(\hat{B})[\hat{\rho}] = \hat{\rho} \hat{B}
+```
+
+# Fields
+- `R::OpType`: The operator ``\hat{B}`` (right-multiplier) as an `AbstractSciMLOperator`
```
ytdHuang:
```suggestion
- `R`: The operator ``\hat{B}`` (right-multiplier) as an `AbstractSciMLOperator`
```

lib/QuantumToolboxCore/src/qobj/superoperators_scimloperators.jl
```diff
@@ -0,0 +1,138 @@
...
+function SciMLOperators.update_coefficients!(op::SpostSuperOperator, u::AbstractMatrix, p, t)
+    update_coefficients!(op.R, u, p, t)
+    return op
+end
+
+@doc raw"""
+    SprePostSuperOperator{T, LOp <: Union{AbstractMatrix, AbstractSciMLOperator}, ROp <: Union{AbstractMatrix, AbstractSciMLOperator}, CT} <: AbstractSciMLOperator{T}
+
+Represents the superoperator ``\mathcal{O}(\hat{A}, \hat{B})[\hat{\rho}] = \hat{A}\hat{\rho}\hat{B}`` acting on the
+**non-vectorized** density operator matrix ``\hat{\rho}``.
```
ytdHuang: same here, make it one-line and use `density operator` instead?

lib/QuantumToolboxCore/src/qobj/superoperators_scimloperators.jl
```diff
@@ -0,0 +1,138 @@
...
+
+The action is defined as:
+```math
+\mathcal{O}(\hat{A}, \hat{B})[\hat{\rho}] = \hat{A} \hat{\rho} \hat{B}
+```
+
+# Fields
+- `L::LOp`: The operator ``\hat{A}`` (left-multiplier) as an `AbstractSciMLOperator`
+- `R::ROp`: The operator ``\hat{B}`` (right-multiplier) as an `AbstractSciMLOperator`
+- `cache::CT`: A cache matrix for intermediate computation (allocated via [`cache_operator`](@ref))
```
ytdHuang:
```suggestion
- `L`: The operator ``\hat{A}`` (left-multiplier) as an `AbstractSciMLOperator`
- `R`: The operator ``\hat{B}`` (right-multiplier) as an `AbstractSciMLOperator`
- `cache`: A cache matrix for intermediate computation (allocated via [`cache_operator`](@ref))
```

lib/QuantumToolboxCore/src/qobj/superoperators_matrix_form.jl
```diff
@@ -0,0 +1,138 @@
...
+        return new{T, LOp, ROp, CT}(L, R, cache)
+    end
+end
+
+function Base.show(io::IO, L::SprePostSuperOperator)
+    a, b = size(L)
+    return print(io, "SprePostSuperOperator($a × $b)")
+end
+
+Base.size(op::SprePostSuperOperator) = size(op.L)
```
ytdHuang, line 104: The size here is ambiguous:

Mathematically, it can be in different size, for example
- `size(L) == m, n`
- `size(R) == n, m`

In this case, this superoperator maps an `n x n` density matrix to an `m x m` one.

However, this is something we want to avoid in solving dynamics.

How about we add the following check in the constructor of `SprePostSuperOperator`:

```julia
m_L, n_L = size(L)
m_R, n_R = size(R)
(m_L == n_L == m_R == n_R) || throw(ArgumentError("Both left- and right-multiplier should be square matrices and share the same size."))
```

lib/QuantumToolboxCore/src/qobj/superoperators_matrix_form.jl
```diff
@@ -0,0 +1,138 @@
...
+
+# Fields
+- `R::OpType`: The operator ``\hat{B}`` (right-multiplier) as an `AbstractSciMLOperator`
+
+See also [`SprePostSuperOperator`](@ref), [`spost`](@ref).
+"""
+struct SpostSuperOperator{T, OpType <: Union{AbstractMatrix, AbstractSciMLOperator}} <: AbstractSciMLOperator{T}
+    R::OpType
+    function SpostSuperOperator(R::OpType) where {OpType <: Union{AbstractMatrix, AbstractSciMLOperator}}
+        T = eltype(R)
```
ytdHuang, line 29: follow the same discussion in `SprePostSuperOperator`:

how about we also make sure it is a squared matrix here:

```julia
m, n = size(R)
(m == n) || throw(ArgumentError("The right-multiplier should be a square matrix."))
```

lib/QuantumToolboxCore/src/QuantumToolboxCore.jl
```diff
@@ -62,6 +62,7 @@ include("qobj/eigen.jl")
 include("qobj/functions.jl")
 include("qobj/states.jl")
 include("qobj/operators.jl")
+include("qobj/superoperators_scimloperators.jl")
```
ytdHuang: I kind of suggest we name the file as `superoperators_matrix_form` or `superoperators_matrix_form_data`, which are more precise.

lib/QuantumToolboxCore/src/qobj/quantum_object_evo.jl
```diff
@@ -166,6 +168,8 @@ function QuantumObjectEvolution(data::AbstractSciMLOperator; type = Operator(),
             sm = isqrt(size(data, 1))
             sn = isqrt(size(data, 2))
             dims = (((sm,), (sm,)), ((sn,), (sn,)))
+        elseif type isa SuperOperatorMatrixForm
+            dims = ((size(data, 1),), (size(data, 2),))
```
ytdHuang: This part seems to be the same as `Operator`, how about:

```julia
    if dims isa Nothing
        if (type isa Operator) || (type isa SuperOperatorMatrixForm)
            dims = ((size(data, 1),), (size(data, 2),))
        elseif type isa SuperOperator
            sm = isqrt(size(data, 1))
            sn = isqrt(size(data, 2))
            dims = (((sm,), (sm,)), ((sn,), (sn,)))
        end
    end
```

albertomercurio:
Fixed

ytdHuang (approved):

PR #726, opened 2026-06-08
Add `save_states` keyword to `mcsolve` to avoid saving states at final time

TendonFFF:
## Checklist

- [ ] Please read [Contributing to Quantum Toolbox in Julia](https://qutip.org/QuantumToolbox.jl/stable/resources/contributing).
- [ ] Any code changes were done in a way that does not break public API.

## Description
`mcsolve` always assembled and (when `keep_runs_results = Val(false)`) trajectory-averaged the per-trajectory states after the trajectories finished. The averaging converts every saved ket to a density matrix via `ket2dm` (an O(dim^2) outer product) for all `ntraj * ntimes` states on a single thread, which can dominate the post-progress-bar finalization time even when the caller only wants expectation values.

Add a `save_states::Union{Val,Bool} = Val(true)` keyword that, when set to `Val(false)`, skips the state assembly/averaging entirely (`sol.states` is left empty). Expectation values are unaffected since they are computed by a separate saving callback. Dispatch is done through a `Val` so the path stays type stable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Add a convenient method when the Hilbert space dimension of all sites are equal.
   - Improve the documentation and docstring.
 - Efficient Jordan Wigner transformation for `fdestroy` and `fcreate`. ([#723])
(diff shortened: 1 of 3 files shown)


ytdHuang:
Hmm

I think this can be done by a keyword argument `save_end` provided by `SciML`.

ytdHuang:
Ok, I made PR #728 to handle this issue.

PR #576, opened 2025-10-26
Add support for arbitrary precision computations

albertomercurio:
## Checklist
Thank you for contributing to `QuantumToolbox.jl`! Please make sure you have finished the following tasks before opening the PR.

- [x] Please read [Contributing to Quantum Toolbox in Julia](https://qutip.org/QuantumToolbox.jl/stable/resources/contributing).
- [x] Any code changes were done in a way that does not break public API.
- [x] Appropriate tests were added and tested locally by running: `make test`.
- [x] Any code changes should be `julia` formatted by running: `make format`.
- [x] All documents (in `docs/` folder) related to code changes were updated and able to build locally by running: `make docs`.
- [x] (If necessary) the `CHANGELOG.md` should be updated (regarding to the code changes) and built by running: `make changelog`.

Request for a review after you have completed all the tasks. If you have not finished them all, you can also open a [Draft Pull Request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) to let the others know this on-going work.

## Description
In this PR, I add the support for arbitrary precision computations, for example using arrays of `BigFloat` type or similar. This allows to do simulations when values smaller than $10^{-14}$ are involved in the computation.

I have tested it to the following use cases:
- Schrodinger equation (`sesolve`)
- Master equation (`mesolve`)
- Sparse eigenvalue problem (`eigsolve(...)` or `eigenstates(A, sparse=true...)`)

For the eigenvalues, the user need to import a package that defines a method for te following functions
- `schur!`
- `ordschur!`

I found [GenericSchur.jl](https://github.com/RalphAS/GenericSchur.jl) to work well, even without writing an extension.

In the next PRs I will write some documentation, as I think it is a very important feature that other packages don't have.

PR #745, opened 2026-07-17
Add documentation for arbitrary precision computations

albertomercurio:
Arbitrary precision support was added in #576, but it was never documented beyond a single changelog line. This adds a Users Guide page for it, plus README and landing page entries.

### The page

- How to use it: the optional element type argument (`destroy(Complex{BigFloat}, N)`), `setprecision`, and eltype inference from values.
- Which extra packages are needed and why — `GenericSchur.jl` for the generic Schur decomposition and `Sparspak.jl` for the sparse shift-and-invert path. These were previously undocumented, and forgetting them gives a bare `MethodError` from `LinearAlgebra`.
- A double-well example: the tunneling splitting falls to ~1e-23, while `Float64` saturates at ~1e-13 and flattens into a noise floor, i.e. it wrongly reports that tunneling stops slowing down. `Double64` tracks `BigFloat` to several digits at a tenth of the cost.
- Caveats: `BigFloat` is slow, no GPU, `eigsolve`'s `tol` default is a hardcoded `Float64` literal, and basis truncation can dominate (at `N = 150` even `BigFloat` gives the wrong splitting).

The page also argues *why* Julia can do this at all — the same generic source respecialized per number type — which QuTiP (hardcoded `double complex`) and JAX (XLA hardware float types only) structurally cannot.

### Bug fix

While writing the examples I found that the `e_ops` expectation value buffer was hardcoded to `ComplexF64`, so `sol.expect` was silently narrowed to double precision even for a `Complex{BigFloat}` problem. It now follows the element type of the problem, in `sesolve`, `mesolve`, `mcsolve`, `ssesolve`, and `smesolve` (the stochastic measurement buffers had the same issue with `Float64`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/qobj/functions.jl
```diff
@@ -75,24 +75,8 @@ function expect(O::AbstractVector{<:AbstractQuantumObject{Operator}}, ρ::Quantu
...
-    return result
-end
 function expect(O::AbstractVector{<:AbstractQuantumObject{Operator}}, ρ::AbstractVector{<:QuantumObject})
-    N_ops = length(O)
-    result = Matrix{ComplexF64}(undef, N_ops, length(ρ))
-    for i in 1:N_ops
-        result[i, :] .= expect.(Ref(O[i]), ρ)
-    end
-    return result
+    return expect.(reshape(O, :, 1), reshape(ρ, 1, :))
```
ytdHuang: would it be better if we just do:


```suggestion
    return expect.(O, transpose(ρ))
```

docs/src/index.md
```diff
@@ -37,6 +37,10 @@ hero:
 
 
 features:
+  - icon: <img width="64" height="64" src="https://img.icons8.com/color/96/atom.png" />
```
ytdHuang: How about we just use QuTiP's logo as the icon here:

```
src="https://qutip.org/images/logo.png"
```

This is qutip's logo link with transparent background

docs/src/users_guide/arbitrary_precision.md
```diff
@@ -0,0 +1,204 @@
...
+    println("$label   naive: ", rpad(Float64(naive(x)), 24), " (rel. err. ", err(naive(x)), ")")
+    println("$label  stable: ", rpad(Float64(stable(x)), 24), " (rel. err. ", err(stable(x)), ")")
+end
+```
+
+`Float64` with the naive formula returns exactly zero: not an inaccurate answer, but a complete loss of every significant digit. Note the two independent levers here — a better *algorithm* (the stable formula) and a better *number type*. Extra precision is not a substitute for numerical care, and the rest of this page is about the cases where care alone is not enough.
+
+## [Using arbitrary precision in QuantumToolbox](@id doc:Arbitrary-Precision:usage)
+
+Every state and operator generating function takes an optional leading type argument:
```
ytdHuang: I suggest:


```suggestion
Every state and operator generating function takes an optional leading argument for specifying element type:
```

docs/src/users_guide/arbitrary_precision.md
```diff
@@ -0,0 +1,204 @@
...
+
+tlist = range(0, 10, 100)
+sol = mesolve(H_big, ψ0_big, tlist, c_ops_big, e_ops = [a_big' * a_big], progress_bar = Val(false))
+
+eltype(sol.expect)
+```
+
+Note that the expectation values come back in the element type you asked for, rather than being silently narrowed to `ComplexF64`.
+
+Be aware of what this does and does not buy you, though. Extra precision does **not** generally make a time-evolution result better: the error of an ODE integration is dominated by the solver tolerances, not by roundoff, and tightening `abstol`/`reltol` in `Float64` is both cheaper and more effective. High-precision solvers earn their cost when the evolution feeds something else that is precision-critical — for example [`eigsolve_al`](@ref), which integrates the master equation to extract Liouvillian eigenvalues, and so inherits exactly the sensitivity described above.
```
ytdHuang:
```suggestion
Be aware of what this does and does not buy you. Extra precision does **not** generally make a time-evolution result better: the error of an ODE integration is dominated by the solver tolerances, not by roundoff, and tightening `abstol`/`reltol` in `Float64` is both cheaper and more effective. High-precision solvers earn their cost when the evolution feeds something else that is precision-critical — for example [`eigsolve_al`](@ref), which integrates the master equation to extract Liouvillian eigenvalues, and so inherits exactly the sensitivity described above.
```

docs/src/index.md
albertomercurio: Done

docs/src/users_guide/arbitrary_precision.md
albertomercurio: Done

docs/src/users_guide/arbitrary_precision.md
albertomercurio: Done

src/qobj/functions.jl
albertomercurio: Done

src/qobj/functions.jl
```diff
@@ -75,24 +75,8 @@ function expect(O::AbstractVector{<:AbstractQuantumObject{Operator}}, ρ::Quantu
...
-    return result
-end
 function expect(O::AbstractVector{<:AbstractQuantumObject{Operator}}, ρ::AbstractVector{<:QuantumObject})
-    N_ops = length(O)
-    result = Matrix{ComplexF64}(undef, N_ops, length(ρ))
-    for i in 1:N_ops
-        result[i, :] .= expect.(Ref(O[i]), ρ)
-    end
-    return result
+    return expect.(O, transpose(ρ))
```
ytdHuang: Hmm, the test is failing.

Seems like `transpose` doesn't work in this way ?

maybe:

```julia
return expect.(O, reshape(ρ, 1, :))
```

albertomercurio: Good catch — `transpose(ρ)` lazily calls `transpose` on each element when indexed, and `QuantumObject` doesn't define that method, so it errors. Pushed `reshape(ρ, 1, :)` instead in ba84904, which reshapes without touching the elements.

---
_Generated by [Claude Code](https://claude.ai/code)_

ytdHuang (approved):

PR #747, opened 2026-07-18
Fix sesolve_map/mesolve_map sharing callback state across trajectories

### LINKED ISSUE #645: Extra callback during mapped problem set up
### Bug Description

Parallel solves using expectation values fail because there is an initialization call to the callback during set up. Consider:
```Julia
H = QobjEvo(sigmax(), (p, t) -> p[1])
tlist = LinRange(0, 1, 100)
ψ0 = basis(2, 0)
```

This works as expected:
```Julia
sol = sesolve(H, ψ0, tlist, params = [π/2], e_ops=(ψ0 * ψ0',))
```

But this fails:
```Julia
prob = sesolveProblem(H, ψ0, tlist, e_ops=[ψ0 * ψ0',])
parameters = [π/2, π/3]
prob_func(prob, i, repeat) = remake(prob, p=[parameters[i]])
sols = sesolve_map(prob, eachindex(parameters), prob_func=prob_func)
```
with an off-by-one error:
```
BoundsError: attempt to access 1×100 Matrix{ComplexF64} at index [1:1, 101]
```

### Code to Reproduce the Bug

```julia

```

### Code Output

```shell

```

### Expected Behaviour

`sesolve_map` should return the expectation values for the times (not times plus one).

### Your Environment

```shell
Package information:
====================================
Julia              Ver. 1.12.1
QuantumToolbox     Ver. 0.40.0
SciMLOperators     Ver. 1.14.1
LinearSolve        Ver. 3.54.0
OrdinaryDiffEqCore Ver. 2.1.0
```

### Additional Context

_No response_

albertomercurio:
The advanced-usage methods of sesolve_map/mesolve_map let callers supply
their own prob_func for the underlying EnsembleProblem, built with
safetycopy=false for performance. If a custom prob_func doesn't itself
reset the e_ops-saving callback's mutable state (the save-index counter
and preallocated expvals matrix), every trajectory ends up aliasing the
same objects, causing a BoundsError or corrupted results once the shared
counter runs past length(tlist).

Add a safetycopy keyword to both advanced-usage methods that defaults to
false when relying on the built-in prob_func (already safe) and true when
a custom prob_func is supplied, while always respecting an explicit
override. This reuses SciMLBase's own deep-copy mechanism instead of
introducing custom copy logic, so a naive custom prob_func is protected
by default without paying the cost for the common/default path.

Fixes #645 

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Ra9uwhCJKFpU1XckVutzSJ


CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/qutip/QuantumToolbox.jl/tree/main)
 
-
```
ytdHuang: Can you also add `[#747]` in the changelog ?

src/time_evolution/mesolve.jl
```diff
@@ -345,6 +345,11 @@ mesolve_map(
 # User can define their own iterator structure, prob_func and output_func
 #   - `prob_func`: Function to use for generating the ODEProblem.
 #   - `output_func`: a `Tuple` containing the `Function` to use for generating the output of a single trajectory, the (optional) `Progress` object, and the (optional) `RemoteChannel` object.
+#   - `safetycopy`: Whether to deep copy the problem before generating each trajectory. Defaults to `false`
+#     when using the built-in `prob_func` (already safe), and to `true` when a custom `prob_func` is supplied,
+#     since a custom `prob_func` that doesn't independently reset per-trajectory callback state (e.g. the e_ops
+#     save counter) would otherwise alias that state across trajectories (see issue #645). Pass `safetycopy =
+#     false` explicitly to opt back into the faster path with a custom `prob_func`, at your own risk.
```
ytdHuang: Can you make this comment just one line ?

```suggestion
#   - `safetycopy`: Whether to deep copy the problem before generating each trajectory. Defaults to `false` when using the built-in `prob_func` (already safe), and to `true` when a custom `prob_func` is supplied, since a custom `prob_func` that doesn't independently reset per-trajectory callback state (e.g. the e_ops save counter) would otherwise alias that state across trajectories (see issue #645). Pass `safetycopy = false` explicitly to opt back into the faster path with a custom `prob_func`, at your own risk.
```

src/time_evolution/sesolve.jl
```diff
@@ -262,6 +262,11 @@ sesolve_map(
 # User can define their own iterator structure, prob_func and output_func
 #   - `prob_func`: Function to use for generating the ODEProblem.
 #   - `output_func`: a `Tuple` containing the `Function` to use for generating the output of a single trajectory, the (optional) `Progress` object, and the (optional) `RemoteChannel` object.
+#   - `safetycopy`: Whether to deep copy the problem before generating each trajectory. Defaults to `false`
+#     when using the built-in `prob_func` (already safe), and to `true` when a custom `prob_func` is supplied,
+#     since a custom `prob_func` that doesn't independently reset per-trajectory callback state (e.g. the e_ops
+#     save counter) would otherwise alias that state across trajectories (see issue #645). Pass `safetycopy =
+#     false` explicitly to opt back into the faster path with a custom `prob_func`, at your own risk.
```
ytdHuang: same here

CHANGELOG.md
albertomercurio: Done

src/time_evolution/mesolve.jl
albertomercurio: Done

src/time_evolution/sesolve.jl
albertomercurio: Done

ytdHuang (approved):

PR #748, opened 2026-07-28
Only force `tstops = tlist` in `ssesolve`/`smesolve` when storing measurements

albertomercurio:
### Problem

The number of solver steps in `ssesolve`/`smesolve` scaled with `length(tlist)` whenever `e_ops` (or a progress bar) was given — 5.5x more steps at 501 output points than at 100, for identical physics. `mesolve` was unaffected.

### Cause

`_generate_stochastic_kwargs` forced `tstops = tlist` for *every* callback. Those tstops exist only for `_homodyne_dWdt!`, which assumes the noise process has a sample at each `tlist` point — i.e. only when `store_measurement = true`. Expectation values are recorded by the interpolating `FunctionCallingCallback`, exactly as in `mesolve`, and never needed them.

Dense tstops are expensive for an adaptive SDE solver: tstop clipping destroys the proposed `dt`, and `SRIW1` needs ~9 steps per interval to grow it back. This reproduces upstream with no QuantumToolbox involved — a plain linear `SDEProblem` goes from 349 to 4469 steps purely from `tstops = range(0, 10, 501)`.

### Fix

Gate the forced tstops on `store_measurement`. `_merge_tstops`'s `prob_is_const` argument is renamed to `skip_tstops`, since it now carries two meanings; the four `isconstant(...)` call sites are unchanged. Time-dependent Hamiltonians still get `tstops = tlist` as before.

Also corrects the documented default SDE tolerances (`reltol=1e-2`/`abstol=1e-2` → the actual `2e-3`/`1e-3`).

### Result

Steps for the reporter's MWE (N=50), median over 10 trajectories:

| npoints | before | after |
|---|---|---|
| 100 | 979 | 820 |
| 251 | 2244 | 746 |
| 501 | 4341 | 779 |

Accuracy is unchanged: against the exact `mesolve` solution over 800 trajectories, the max deviation is 0.060 / 0.065 / 0.069 at 101 / 251 / 501 points (peak ⟨n⟩ = 9.5) — flat, so the interpolated recording introduces no resolution-dependent bias. `store_measurement = true` keeps its tstops and its results are unchanged.

Full test suite passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -555,4 +555,3 @@ Release date: 2024-11-13
 [#745]: https://github.com/qutip/QuantumToolbox.jl/issues/745
 [#747]: https://github.com/qutip/QuantumToolbox.jl/issues/747
 [#748]: https://github.com/qutip/QuantumToolbox.jl/issues/748
-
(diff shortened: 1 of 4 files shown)


ytdHuang (approved):
LGTM!

by the way, nice catch about the incorrect `reltol` and `abstol` in the docstring from Claude code lol

albertomercurio:
Ahahah

